PR #2, opened 2024-08-26
TIMX 338 - Add init-job functionality

ghukill:

The first step in starting a Transmogrifier A/B test is creating a "Job" which is a wrapper around subsequent "Runs" of the job.  The job provides a working directory, later the Transmogrifier versions used, etc.  This information is captured in a `job.json` file that lives in the Job's working directory.

The first function importable from `abdiff.core` has been created to support this: `init_job()`.  It takes only a single argument `job_name`, and the rest is opinionated  behavior from `abdiff.core.utils`, where it is intended that other functions that require knowing a job name, slug, or working directory can use those as well.

Start an ipython shell
```shell
pipenv run ipython
```

Create a new job:
```python
from abdiff.core import init_job

job_data = init_job("Fruits and Vegetables")

display(job_data)
#Out[3]: 
#{'job_name': 'Fruits and Vegetables',
# 'job_slug': 'fruits-and-vegetables',
# 'working_directory': 'output/fruits-and-vegetables'}
```

If no environemnt variables are set, you should see a new directory in the local `output` folder:
```
/ output
├── fruits-and-vegetables
│   └── job.json
```

This is basically the totality of this function's requirements.  Any downstream jobs can use `utils.get_job_slug_and_working_directory()` to get a job's slug and working directory from the job name alone.

Delete this section if it isn't applicable to the PR.

YES

NO

- https://mitlibraries.atlassian.net/browse/TIMX-338

### Developer

### Code Reviewer(s)


abdiff/core/__init__.py
```diff
@@ -2,3 +2,9 @@
 
 All primary functions used by CLI are importable from here.
 """
+
+from abdiff.core.init_job import init_job
+
+__all__ = [
+    "init_job",
+]
```
ghukill, line 10: I had figured we could just add functions here as they get created?

abdiff/core/utils.py
```diff
@@ -0,0 +1,38 @@
...
+CONFIG = Config()
+
+
+def get_job_slug_and_working_directory(job_name: str) -> tuple[str, Path]:
+    """Create working directory for new job by slugifying job name."""
+    job_slug = slugify(job_name)
+    return job_slug, Path(CONFIG.data_directory) / job_slug
+
+
+def update_or_create_job_json(job_name: str, new_job_data: dict) -> dict:
```
ghukill, line 20: To anyone who works' on functionality that will update the job's `job.json`, my hope is that this utility function would be reusable for that purpose.

abdiff/config.py
```diff
@@ -1,4 +1,22 @@
...
+    def __getattr__(self, name: str) -> Any:  # noqa: ANN401
+        """Method to raise exception if required env vars not set."""
+        if name in self.REQUIRED_ENV_VARS or name in self.OPTIONAL_ENV_VARS:
+            return os.getenv(name)
+        message = f"'{name}' not a valid configuration variable"
+        raise AttributeError(message)
+
+    @property
+    def data_directory(self) -> str:
+        return self.DATA_DIRECTORY or "output"
```
ehanson8, line 19: Could this just be `output_directory` for simplicity?

abdiff/core/init_job.py
```diff
@@ -1 +1,36 @@
...
+logger = logging.getLogger(__name__)
+
+
+def init_job(job_name: str) -> dict:
+    """Function to initialize a new Job.
+
+    1. create a working directory for job
+    2. initialize a job.json file
+    """
+    job_slug, job_working_directory = get_job_slug_and_working_directory(job_name)
```
ehanson8, line 23: Building on the `output_directory` comment, to further minimize potential confusion, could this just be `job_directory` or `working_directory` consistently?

abdiff/core/init_job.py
```diff
@@ -1 +1,36 @@
...
+    os.makedirs(job_working_directory)
+    logger.info(
+        f"Job '{job_slug}' initialized.  Working directory: {job_working_directory}"
+    )
+
+    job_data = {
+        "job_name": job_name,
```
ehanson8, line 30: How do you picture this getting used later? It feels like the slug could possibly suffice

ehanson8 (commented):
Looks good but a few comments about directory naming

abdiff/core/init_job.py
jonavellecuerdo: I agree with @ehanson8 on choosing one naming convention and keeping it consistent across functions.
 
I am leaning towards `job_directory`. I like that it is explicitly linking the directory to a job.  Related to [@ehanson8 's comment on `output_directory`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/2/files#r1731591966): `data_directory` makes me think of a folder containing file extracts and transformed record JSON files, but this directory will contain more than that -- the structure of the directory is very specific to a job, containing job and run JSON files and dated 'run' folders _in addition to_ the transformed records. 🤔

abdiff/config.py
ghukill: Maybe pulling in [your comment below](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/2#discussion_r1731592951) @ehanson8: what if we named the property `root_working_directory`?  and updated the env var to `ROOT_WORKING_DIRECTORY`?  

Then we'd have the following throughout:
- `root_working_directory`: the base folder for Jobs to get created in
- `job_working_directory`: a Job specific folder, under the root working directory

abdiff/core/init_job.py
ehanson8: That's fine to leave it in, just asking the question because I could see it becoming irrelevant in the future depending on the flow of the app. So I will re-raise if that happens!

abdiff/config.py
ehanson8: That works for me, just looking for symmetry and alignment between env var, args, and app variables whenever possible!

jonavellecuerdo: Saw your comment after I wrote [my input](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/2#discussion_r1731650388), but I like what you've proposed above -- `root_working_directory` and `job_working_directory`.

abdiff/core/init_job.py
jonavellecuerdo: I thought the use of `slugify` was nifty and a nice way to manage "human-readable" job names while also supporting neat, consistent filepaths. 🤔

abdiff/config.py
ghukill: Thanks @ehanson8 and @jonavellecuerdo for the feedback. I agree it was a little awkward before.  Working on some changes now and will re-request a review.

ghukill:
@ehanson8, @jonavellecuerdo - I've renamed `data_directory` to `root_working_directory` throughout in this commit: https://github.com/MITLibraries/transmogrifier-ab-diff/pull/2/commits/b51e8fdfc70998196e08c084900e438b82b44426.

While we're at it: I wonder if the default directory should be `/tmp` instead of `./output` relative to the project directory?

At least in my IDE, I have to be careful that it doesn't attempt to scan and index that directory because it's part of the project.

Pushing another commit now that will change the default to `/tmp`.  Will re-request review shortly.

Lastly, while I imagine we may need to tune some things as it comes together more, we really haven't discussed the potentially fairly large data this app will be saving locally.  Probably only hundreds of megabytes... unless someone gets ambitious and tries to run on full Alma (20x files @ 1gb each).  Counter-point, this is naturally pretty hard to do by accident given how slow the transform is for those files; it would be a potentilaly multi-day running job locally.

Somewhat unrelated to the default root working directory, but felt worth mentioning as we're thinking about files on local disk.

ghukill:
> While we're at it: I wonder if the default directory should be `/tmp` instead of `./output` relative to the project directory?
> 
> At least in my IDE, I have to be careful that it doesn't attempt to scan and index that directory because it's part of the project.
> 
> Pushing another commit now that will change the default to `/tmp`. Will re-request review shortly.

Scratch that!

Going to create a ticket to think about file management over time, job footprints, etc.  Seems conceivable we might even want to support the job working directory being in S3.  For now, feels okay to have the root working directory be `<repository/output`.

ghukill:
> > While we're at it: I wonder if the default directory should be `/tmp` instead of `./output` relative to the project directory?
> > At least in my IDE, I have to be careful that it doesn't attempt to scan and index that directory because it's part of the project.
> > Pushing another commit now that will change the default to `/tmp`. Will re-request review shortly.
> 
> Scratch that!
> 
> Going to create a ticket to think about file management over time, job footprints, etc. Seems conceivable we might even want to support the job working directory being in S3. For now, feels okay to have the root working directory be `<repository/output`.

Ticket created: https://mitlibraries.atlassian.net/browse/TIMX-342

jonavellecuerdo:
> Rename data_directory to root_working_directory

@ghukill Curious what you mean by your IDE "attempt[ing] to scan and index" a directory? 🤔

jonavellecuerdo (approved):

ghukill:
> > Rename data_directory to root_working_directory
> 
> @ghukill Curious what you mean by your IDE "attempt[ing] to scan and index" a directory? 🤔

My IDE (PyCharm) indexes all the files in the project.  This is great for searching around code as it makes it quicker... but is less great when it finds 20gb of JSON files and starts to index them all for searching 😅.  It's really not designed for "data" your app may use, just the code.  

I can exclude a directory if I remember, and short-circuit this problem.  But I think it's consistent with a more general concept that storing ephemeral, working data inside a project's sort of "runtime" environment is not a great idea.  For this reason, and others.  

Hoping this ticket, https://mitlibraries.atlassian.net/browse/TIMX-342, can become the locus of these discussions.

ehanson8 (approved):

PR #6, opened 2024-08-30
Timx 341 run ab transforms

jonavellecuerdo:

[DRAFT] Implement function to run transforms using A/B Docker images of `transmogrifier` via Docker containers. This function provides an easy way to programmatically run Docker containers using the [Docker python module (mainly the "Containers" module)](https://docker-py.readthedocs.io/en/stable/containers.html).

`run_ab_transforms` achieves the outlined functionality, function signature, and inputs and outputs in the [Core Functions and CLI signatures wiki](https://github.com/MITLibraries/transmogrifier-ab-diff/wiki/Core-Functions-and-CLI-signatures#run_ab_transforms). As described in the wiki, this function will:

* reads input files
* runs A/B transforms for all files
* creates transformed/a and transformed/b sub-directories under <JOB>/runs/<RUN> directory
* updates run.json with the input files used
* **return** tuple of a/b transformed files: tuple[list, list]
paths are relative to run directory, e.g.

Linting has been applied. **No unit tests yet!**

YES

NO

- https://mitlibraries.atlassian.net/browse/TIMX-341

### Developer
- [ ] All related Jira tickets are linked in commit message(s)
- [ ] Stakeholder approval has been confirmed (or is not needed)

### Code Reviewer(s)


abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,203 @@
...
+    This method will parse command arguments from the provided 'input_file'.
+    The generated transform command arguments are used when running container(s)
+    with the transmogrifier Docker images.
+    """
+    input_file_name = input_file.split("/")[-1]
+    input_file_ext = input_file_name.split(".")[-1]
+    output_file_name = input_file.split("/")[-1].replace(
+        f"extracted-records-to-index.{input_file_ext}",
+        "transformed-records-to-index.json",
+    )
```
ghukill, line 185: If we include the `source` [parsing above](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR52), this is a lot of parsing and constructing based on the file naming convention for TIMDEX files.  

While I like the responsibility of this function to generate commands, perhaps it makes sense to break-out the filename parsing to a standalone function that could be tested?

I recently faced this while working on the parquet exploration repo, using [regex to parse the pieces of the filename](https://github.com/MITLibraries/dataeng-workspace-parquet/blob/main/pqsandbox/timdex/s3.py#L19-L36).  

Not sure it's totally applicable here... as you may not need to know the date or the cadence (as examples), but thought worth sharing.  This regex also handles the edge cases where an source like `alma` adds a `_01`, `_02`, etc., suffix to the filename before the file extension.  Though with what you have, think it would be fine (it'd just get copied).

So maybe... just a function like `get_transform_filename_from_extract_filename`?

ghukill: But maybe over-complicating things!  Feels as though this would work.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,203 @@
...
+            )
+            logger.info(
+                f"Container '{container.id}' (Docker image: {docker_image}) "
+                f"RUNNING transform for '{source}' input_file: {input_file}."
+            )
+            container_ids.append(container.id)
+
+    # wait for all container runs to complete (status = "exited")
+    # get logs from all Docker containers
+    get_logs(run_id, container_ids, log_file=f"{run_directory}/transformed/logs.txt")
```
ghukill, line 79: It feels like aggregating the container logs into a single file is actually part of a more fundamental thing happening here: waiting for all containers to complete.

A lot of parallel processing libraries have methods like `.join()` or `.wait()`.  Might be worth considering if the `get_logs()` function could be renamed / slightly re-documented to be a function that will wait for a list of containers.

Perhaps it could return all the logs from the containers still as python strings?  And then this "main" function could write them to a file; keep it's scope narrow to wait + get logs (which is basically the task's output).

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,203 @@
...
+            for container in exited_containers:
+                logs += str(
+                    write_container_logs(
+                        container,
+                        log_file,
+                    )
+                )
+                logger.info(logs)
+                logged_container_ids.add(container.id)
+        time.sleep(5)
```
ghukill, line 142: Minor, but think this could be much smaller, like `0.25`.  It's just a harmless, quick check to see if something's done.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,203 @@
...
+        # get logs from completed container runs for a given "run"
+        exited_containers = [
+            container
+            for container in DockerClient.containers.list(
+                filters={"label": f"run={run_id}", "status": "exited"}
+            )
+            if container.id in container_ids and container.id not in logged_container_ids
+        ]
+        if any(exited_containers):
+            for container in exited_containers:
```
ghukill, line 133: If I'm understanding, this works nicely because we don't know how many containers finished since last we checked -- could be multiple -- but we sequentially write the logs out.  We avoid a pretty easy pitfall in parallel work where outputs come in from disparate tasks at the same time and jumble up logging or data collection.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,203 @@
...
+        filepaths.append(sub_filepaths)
+    return tuple(filepaths)
+
+
+def get_logs(run_id: str, container_ids: list[str], log_file: str) -> str:
+    """Monitor Docker container runs and write logs to file as containers exit."""
+    logged_container_ids: set[str] = set()
+    logs = ""
+    while logged_container_ids != set(container_ids):
+        if not DockerClient.containers.list(all=True, filters={"label": f"run={run_id}"}):
```
ghukill, line 121: If we know the list of container ids we care about, why do we need to filter by the label `run`?   Would it be possible to just get _all_ containers, and then filter _them_ to only those in the passed container ids?  

And/or, looks like you can filter to the "ancestor", which is the image of the container?

[From docs](https://docker-py.readthedocs.io/en/stable/containers.html?highlight=def%20list#docker.models.containers.ContainerCollection.list):
```
- `ancestor` (str): Filter by container ancestor. Format of
    ``<image-name>[:tag]``, ``<image-id>``, or
    ``<image@digest>``.
```

I ask, as the `run_id` jumped out at me in the main function.  I think it _is_ the run ID for all intents and purposes, but it's kind of the first time we mentioned it as such.  Until now, we've just referred to a run by `run_directory`.  I don't think it's wrong, but just noting this is the first time we kind of elevate the timestamp to an "id" status.

Like.... maybe each run _should_ have an ID?  We could store this in the `run.json`?  Just some thoughts.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,203 @@
...
+    start_time = perf_counter()
+
+    # get run ID (used to label containers)
+    run_id = str(run_directory).split("/")[-1]
+
+    # create directories for 'transformed' outputs
+    transformed_directory_a, transformed_directory_b = create_directories(run_directory)
+
+    # create list to track container IDs for a given 'run'
+    container_ids = []
```
ghukill, line 41: Very minor: I'd kind of expect this right above the loop that is going to utilize it (around line  # 51)

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,203 @@
...
+                    "source": source,
+                    "input_file": input_file,
+                    "run": run_id,
+                },
+                volumes=[f"{transformed_directory}:/tmp"],
+            )
+            logger.info(
+                f"Container '{container.id}' (Docker image: {docker_image}) "
+                f"RUNNING transform for '{source}' input_file: {input_file}."
+            )
```
ghukill, line 74: Maybe worth breaking this out into a standalone function? One that kicks back a running container?

abdiff/config.py
```diff
@@ -4,7 +4,12 @@
 
 
 class Config:
-    REQUIRED_ENV_VARS = ("WORKSPACE",)
+    REQUIRED_ENV_VARS = (
+        "AWS_ACCESS_KEY_ID",
+        "AWS_SECRET_ACCESS_KEY",
+        "AWS_SESSION_TOKEN",
```
ghukill, line 10: We could address this later perhaps, but wondering if there might be other ways to get AWS credentials to the container.

In the past, I've enjoyed/preferred using a mount where I mount my `~/.aws` directory to the container's `/root/.aws` directory.  This was I can use my SSO set credentials without ever getting these tokens, that expire.

Not proposing that way is _better_, but might be worth some thinking / research on ways to have a container have AWS credentials.

ghukill: Or.... what if we download the file to the host machine, and then mount that in the container for use?  This might have an advantage: for any given file the container would be reading from S3, it's going to happen _twice_ given the A/B nature of this application.

If we downloaded locally, we'd only pay the network penalty once for downloading the file to the host machine.  Then we'd mount for the container to use, it transforms from it, and returns a transformed file.  Once all containers are done (or more surgically, when all containers that require _that_ file are done) we could delete from the host machine.  Or heck, we could even have an option to keep it around if we know we're going to re-run some transformations.  

This adds a bit of complexity, but I think it would halve the time we spend on network reading.  And, as mentioned above, it might open the option for not deleting the files we've retrieved if we know we're doing more runs.  I'd happily allow some XML/JSONLines file to accumate on my host machine if I know it would dramatically increase the runtimes for the A/B diffs.  Happy to discuss more.

We could also make a ticket for this, and just keep as-is.  But FWIW, this would remove the need to pass AWS credentials to the container, which has advantages as well.

abdiff/core/run_ab_transforms.py
ehanson8: Agreed!

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,203 @@
...
+            container_ids.append(container.id)
+
+    # wait for all container runs to complete (status = "exited")
+    # get logs from all Docker containers
+    get_logs(run_id, container_ids, log_file=f"{run_directory}/transformed/logs.txt")
+
+    # clone run data and add list of input files
+    run_data = read_run_json(run_directory)
+    run_data.update({"input_files": input_files})
+    update_or_create_run_json(run_directory, run_data)
```
ehanson8, line 84: Since you're just adding 1 key/value pair, I think you could just do `update_or_create_run_json(run_directory, {"input_files": input_files})` without the previous 2 lines

abdiff/core/run_ab_transforms.py
ehanson8: Good points, I'm all in favor of minimizing the "IDs" we have to keep track of so if possible sticking with the `run_directory` to ID the run

abdiff/config.py
ghukill: On the flipside, perhaps it would make sense to revist later?  I see two things I'd like this application to be able to do eventually, which require some tweaks to this step:

1. read input files from "local" host machine, relative to the docker containers
2. run ab-diff on _two, distinct_ inputs sets of files_ for _one_ version of Transmog

abdiff/core/run_ab_transforms.py
ghukill: For the record, I might propose that we do create unique IDs for each run, like a UUID.  We can store this in the run JSON, and maybe we'll never need it, but it would be a way to absolutely unambiguously reference every run within a job.

abdiff/core/__init__.py
```diff
@@ -5,5 +5,6 @@
 
 from abdiff.core.init_job import init_job
 from abdiff.core.init_run import init_run
+from abdiff.core.run_ab_transforms import run_ab_transforms
 
-__all__ = ["init_job", "init_run"]
+__all__ = ["init_job", "init_run", "run_ab_transforms"]
```
ghukill, line 10: Missed in my first pass, but +1 to this pattern!  I think it will be nice to do `from abdiff.core import XYZ` for all the main core functions.

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Good idea -- and will remember this suggestion moving forward! Updated.

abdiff/core/run_ab_transforms.py
jonavellecuerdo: I agree that a run UUID would be helpful! While writing / testing the code above, I was executing a lot of Docker containers, and when I looked at Docker Desktop, I realized I had a long list of exited Docker containers and didn't have a way to connect the executed Docker container with a specific run. 🤔 This led to the decision to label the Docker containers with a "run ID", which -- as the code is currently written -- would translate to the lowest level of the run directory: the run timestamp (e.g., `2024-09-05_11-42-00`).

The first `if` condition may not be necessary, but I do think filtering by the "run ID" is a requirement as _all_ exited Docker containers (including containers from other runs) will be returned otherwise. The result will also be the same if we filter by ancestor (i.e., one docker image can be associated with many runs) Let me know what you think given this additional information!

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Let me know what you think of the [`run_docker_container` function](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR71-R104). Note the removal of the `generate_transform_command`; after reworking the parsing method, the remaining contents of the `generate_transform_command` felt minimal / not requiring its own function.

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Indeed! When retrieving the "exited" containers, we are making sure to exclude exited containers that have already been _logged_ (i.e., `container.id not in logged_container_ids`) to avoid duplication and the structure of the for loop results in the logs writing to the file as they finish!

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Updated!

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Please see the added function [`parse_transform_details_from_extract_filename`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR187-R202). Thank you for sharing the parsing method with regex! I ultimately went with the regex approach as it felt more robust but simplified the regex pattern for this use case.

abdiff/core/run_ab_transforms.py
ghukill: Trying to be succinct here, and summarize the discussion above: minting a run UUID makes sense for the reasons you outline @jonavellecuerdo.  When we start a run, we know we're going to start `n` containers, and we know we want to wait for _only_ those containers to finish.  Therefore, if we know each container will have a tag/label of the run UUID, that feels like an airtight way to filter down to only those containers as ones we're interested in.

But, I'd be remiss if I didn't ask a followup question that maybe got buried above.

[Right here](https://github.com/MITLibraries/transmogrifier-ab-diff/blob/TIMX-341-run-ab-transforms/abdiff/core/run_ab_transforms.py#L52), when we kickoff a container, it feels like we have the _specific_ container in hand in code.  If we collect those in a list, couldn't we just loop through those containers and see when they're done?  Or maybe I'm thinking about that wrong, that perhaps they don't provide a `.done()`, `.finished()`, `.still_running` method or something?

If we use the actual container instances that we have in python memory, I suppose I wonder why we'd need to query the `docker_client` for containers, as we already have them.

abdiff/config.py
jonavellecuerdo: As discussed via Slack huddle, it was decided to **leave the method for passing AWS credentials as-is** -- creating a `.env` file -> copy-pasting credentials -> relying on `Config` to load environment variables as needed in the call to `run_ab_transforms`. While the A/B transform workflow will be run locally, minimizing security risks, it's generally recommended to avoid mounting files with sensitive info into Docker containers.

As for the latter comments on updating method for accessing files, it was decided to proceed with the current approach of transforming input files from S3 as a "first pass" version of this command. Ticket [TIMX-353](https://mitlibraries.atlassian.net/browse/TIMX-353) has been created to explore this topic further.

abdiff/core/run_ab_transforms.py
ghukill: Agreed!  The `run_docker_container` function looks great; one purpose, easy to follow.  Will wait for the "real" PR to dig into how and when it's called, but doesn't feel like it needs any further refactoring.

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Ah! That is a good point. Instead of collecting container IDs, collect the Container objects themselves -- taking a more object-oriented approach! Let me ponder this and see what changes can be made from this!

ghukill: Mentioned in slack, but noting here as well: I think the collected `Container` objects do _not_ automatically update their status, health, etc.  But I believe you can call `Container.reload()` which will update things like `.status`.

abdiff/config.py
ghukill: Sounds good!  Thanks for considering and the summary.

abdiff/core/run_ab_transforms.py
jonavellecuerdo: See https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6#discussion_r1747096635.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,183 @@
 """abdiff.core.run_ab_transforms"""
+
+import logging
+import os
+import re
+import sys
+
+sys.path.append("/Users/jcuerdo/Documents/repos/transmogrifier-ab-diff")
```
ghukill, line 8: I reckon just a temporary shim for testing, but this should get removed.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,183 @@
...
+            "AWS_SECRET_ACCESS_KEY": CONFIG.AWS_SECRET_ACCESS_KEY,
+            "AWS_SESSION_TOKEN": CONFIG.AWS_SESSION_TOKEN,
+        },
+        labels={
+            "docker_image": docker_image,
+            "source": source,
+            "input_file": input_file,
+            "run": run_id,
+        },
+        volumes=[f"{transformed_directory}:/tmp"],
```
ghukill, line 113: Happy to wait for the "real" PR to dig into this if you'd like, but in some local testing I _think_ this has to be an absolute path to mount into the container.  

We could note this somewhere and make it the responsibility of the calling context to _always_ pass an absolute directory for the `run_directory` to the parent `run_ab_transforms` function. In this way, we could be sure that the calculated transformed directory(s) are absolute.

Or -- don't know offhand -- maybe python has a way to take a relative directory, and get the absolute path from that?

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,183 @@
...
+        for container in containers:
+            header = (
+                f"docker_image: {container.labels['docker_image']} | "
+                f"source: {container.labels['source']} | "
+                f"input_file: {container.labels['input_file']}\n"
+            )
+            file.write(header.encode())
+            file.writelines(container.logs(stream=True))
+            file.write(b"\n")
+    logger.info(f"Log file created: {log_file}")
```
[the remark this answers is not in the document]
jonavellecuerdo: Yes, they are returned as bytes! I've[ simplified the file-writing method](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR70-R72).

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Resolving this comment for now! Will continue to explore if there are better ways to simplify the code for waiting + log retrieval.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+logger = logging.getLogger(__name__)
+
+
+def run_ab_transforms(
+    run_directory: str,
+    image_name_a: str,
+    image_name_b: str,
+    input_files: list[str],
+    docker_client: docker.client.DockerClient | None = None,
+    timeout: int = 180,
```
jonavellecuerdo, line 29: The value for `timeout` gets used by the [`wait_for_containers` function](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR133-R136).

tests/conftest.py
```diff
@@ -31,6 +108,36 @@ def job(job_directory):
...
+    transformed_directory_a = str(Path(run_directory) / "transformed/a")
+    transformed_directory_b = str(Path(run_directory) / "transformed/b")
+    return transformed_directory_a, transformed_directory_b
+
+
+@pytest.fixture
+def create_transformed_directories(run_directory):
+    return create_directories(
+        base_directory=run_directory, subdirectories=["transformed/a", "transformed/b"]
+    )
```
jonavellecuerdo, line 138: It's worth nothing that these two fixtures are different. 

* `transformed_directories`:  only returns a tuple of strings representing transformed directories A and B.
   * This is used by mocks used by the test for `run_ab_transforms` because the `run_ab_transforms` will create the directories.
   
* `create_transformed_directories`: returns a tuple of strings representing transformed directories A and B _**and**_ actually creates the transformed directories. 
   * This is used by tests for `run_docker_container` and `get_transformed_files`.

tests/conftest.py
```diff
@@ -45,3 +152,47 @@ def mocked_docker_client():
...
+        "transmogrifier-example-job-1-def456:latest",
+    ]
+)
+def mocked_docker_container_and_image(
+    request, mocked_docker_container_a, mocked_docker_container_b
+):
+    if "abc123" in request.param:
+        yield mocked_docker_container_a, request.param
+    elif "def456" in request.param:
+        yield mocked_docker_container_b, request.param
```
jonavellecuerdo, line 169: See docstring in the test for [`run_docker_container_success`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-ef381e309d69d9280a44bcd8e1f647b38878fd2167b92011767f6f351ce4363dR52-R89).

tests/conftest.py
```diff
@@ -31,6 +108,36 @@ def job(job_directory):
...
+    transformed_directory_a = str(Path(run_directory) / "transformed/a")
+    transformed_directory_b = str(Path(run_directory) / "transformed/b")
+    return transformed_directory_a, transformed_directory_b
+
+
+@pytest.fixture
+def create_transformed_directories(run_directory):
+    return create_subdirectories(
+        base_directory=run_directory, subdirectories=["transformed/a", "transformed/b"]
+    )
```
jonavellecuerdo, line 138: It's worth nothing that these two fixtures are different. 

* `transformed_directories`:  only returns a tuple of strings representing transformed directories A and B.
   * This is used by mocks used by the test for `run_ab_transforms` because the `run_ab_transforms` will create the directories when called.
   
* `create_transformed_directories`: returns a tuple of strings representing transformed directories A and B _**and**_ actually creates the transformed directories. 
   * This is used by tests for `run_docker_container` and `get_transformed_files`.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+) -> tuple[list[str], ...]:
+    """Run Docker containers with versioned images of Transmogrifier."""
```
ghukill, line 31: I think a very _high_ level flow added to this docstring could be helpful, something along the lines of:

```
1. Directories are created to capture the transformed files
2. For all input files, an A and B version of Transmogrifier is run via detached Docker containers, resulting in all containers running in parallel
3. Wait for all containers to complete
4. Aggregate logs from all containers
5. Return a list of all A and B transformed files
```

To the degree this is added, it may be that some comments in code then feel unneeded or redundant (but definitely keep if they feel helpful!)

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+            )
+            containers.append(container)
+
+    # wait for all container runs to complete (status = "exited")
+    wait_for_containers(containers, timeout)
+    logger.info(f"All {len(containers)} containers have exited.")
+
+    # get logs from all Docker containers
+    log_file = f"{run_directory}/transformed/logs.txt"
+    with open(Path(log_file), "w") as file:
```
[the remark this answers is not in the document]
ehanson8: I see it's used all over, so same question for all occurrences!

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+    exited_containers: list[Container] = []
+    start_time = time.time()
+    while len(exited_containers) < len(containers):
+        running_containers = [
+            container for container in containers if container not in exited_containers
+        ]
+        if time.time() - start_time >= timeout:
+            raise DockerContainerRuntimeExceededTimeoutError(
+                containers=containers, timeout=timeout
+            )
```
ghukill, line 136: Nice addition!  I like this, if I'm understanding correctly, as it will report _all_ containers that have exceeded the timeout.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+        for container in containers:
+            header = (
+                f"docker_image: {container.labels['docker_image']} | "
+                f"source: {container.labels['source']} | "
+                f"input_file: {container.labels['input_file']}\n"
+            )
+            file.write(header)
+            file.write(container.logs().decode())
+            file.write("\n\n")
+    logger.info(f"Log file created: {log_file}")
```
ehanson8, line 78: `run_ab_transforms` is doing so much but overall is very readable! My only suggestion is making this block a separate function

abdiff/core/utils.py
```diff
@@ -48,3 +48,14 @@ def update_or_create_job_json(job_directory: str, new_data: dict) -> dict:
 def update_or_create_run_json(run_directory: str, new_data: dict) -> dict:
     """Create or update a run's JSON file."""
     return update_or_create_json(run_directory, "run.json", new_data)
+
+
+def create_subdirectories(
+    base_directory: str, subdirectories: list[str]
+) -> tuple[str, ...]:
```
ghukill, line 55: I think a tiny docstring here could be helpful, touching on _why_ this is helpful  (i.e. some functions have to create a few subdirectories, and this normalizes an approach).

tests/conftest.py
```diff
@@ -1,9 +1,86 @@
...
+        self.count = 0
+
+    def yield_mocked_run(self, transformed_directories: tuple[str]) -> None:
+        """Perform a mocked run of transmogrifier.
+
+        This function yields a different outcome based on the value of
+        MockedContainerRun.count each time it is called. The function will
+        create a placeholder transformed file in the A/B transformed
+        directories in the correct order (first in 'transformed/a' then
+        in 'transformed/b').
```
ghukill, line 46: Nice docstring here, very helpful to understand why and how this class is used.

tests/conftest.py
```diff
@@ -1,9 +1,86 @@
...
+        """
+        transformed_directory_a, transformed_directory_b = transformed_directories
+        if self.count == 0:
+            self.count += 1
```
ghukill, line 50: Very minor, but `self.count += 1` could be bumped outside of the `if/elif` checks, then bump both of those by 1, e.g.

```python
self.count += 1
if self.count == 1:  # first time called
  # do stuff
elif self.count == 2: # second time called
  # do stuff
else:
  # ruh-roh, shouldn't be here
```

tests/test_run_ab_transforms.py
```diff
@@ -0,0 +1,154 @@
...
+    assert mocked_docker_container_b.status == "exited"
+
+
+def test_wait_for_containers_raise_error(
+    mocked_docker_container_a, mocked_docker_container_b
+):
+    with pytest.raises(DockerContainerRuntimeExceededTimeoutError):
+        wait_for_containers(
+            containers=[mocked_docker_container_a, mocked_docker_container_b],
+            timeout=0,  # force timeout
```
ghukill, line 108: 👍

abdiff/core/exceptions.py
```diff
@@ -0,0 +1,18 @@
...
+        self.timeout = timeout
+        super().__init__(self.get_formatted_message())
+
+    def get_formatted_message(self) -> str:
+        container_ids = [container.id for container in self.containers]
+        return (
+            f"Timeout of {self.timeout} seconds exceeded."
+            f"{len(container_ids)} container(s) is/are still running:"
+            f"{container_ids}."
+        )
```
ghukill, line 18: I dig it!

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+            container = run_docker_container(
+                docker_image, transformed_directory, input_file, docker_client
+            )
+            containers.append(container)
+
+    # wait for all container runs to complete (status = "exited")
+    wait_for_containers(containers, timeout)
+    logger.info(f"All {len(containers)} containers have exited.")
+
+    # get logs from all Docker containers
```
ghukill, line 66: Hate to even ask, but what about breaking this out as a little standalone function as well, where you pass it the list of `containers`?

ghukill (changes requested):
I was waffling between a "Comment" and "Request changes" response, as I think 99% of my comments are just optional and questions.  But I leaned towards "Request changes" if for no other reason than a high level docstring for the main `run_ab_transforms()` function; think it would benefit.

Although I left a handful of comments and optional changes, I think overall this looks great.  Given our out-of-PR discussions, I can appreicate the complexity in testing the docker container behavior.  The final testing approach feels robust and understandable (after some time spent reading), and I think will serve us well as we build on top of this.

ehanson8 (commented):
Great work here, a few questions and suggestions!

abdiff/core/run_ab_transforms.py
ehanson8: Agreed ☝️

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+        dirs.sort()
+        sub_filepaths = []
+        if root.endswith("transformed"):
+            # skip 'transformed' root directory
+            continue
+        for filename in files:
+            relative_dir = root.replace(run_directory, "")
+            sub_filepaths.append(os.path.join(relative_dir, filename))
+        filepaths.append(sub_filepaths)
+    return tuple(filepaths)
```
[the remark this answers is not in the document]
ehanson8: Used `glob` in `dspace-api-python-scripts` up until our S3 conversion @jonavellecuerdo, so you could find examples in the old commits

abdiff/core/run_ab_transforms.py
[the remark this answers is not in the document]
ghukill: Nice catch - while I've been liking the use of `Path` in this project, this feels like an example where it's not needed.  Or, to rework the f-string part into a `Path()` approach and use that.

abdiff/core/run_ab_transforms.py
ghukill: Now that I've read through this closer, I see how `run_ab_transforms()` will pass an absolute path to the `run_docker_container` helper function when it runs.

That said, I did learn this is a thing, which could be an optional safeguard to add to the `run_docker_container()` function to ensure any path that is mounted is absolute:

```python
import os
absolute_path = os.path.abspath(relative_path)
```

abdiff/core/run_ab_transforms.py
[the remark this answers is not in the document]
ghukill: Given the [comment above](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6#discussion_r1755301616) about using `os.path.abspath` I realized there is a `os.path.relpath` as well!

Was curious how it would fit here:
```python
ordered_files = []
for version in ["a", "b"]:
    absolute_filepaths = glob.glob(f"{run_directory}/transformed/{version}/*.json")
    relative_filepaths = [
        os.path.relpath(file, run_directory) for file in absolute_filepaths
    ]
    ordered_files.append(relative_filepaths)
return tuple(ordered_files)
```

Interestingly, this revealed what could be a bug?  The test `XXX` failed:

```
  Full diff:
    (
        [
  -         '/transformed/a/source-2024-01-01-full-transformed-records-to-index.json',
  ?          -
  +         'transformed/a/source-2024-01-01-full-transformed-records-to-index.json',
        ],
        [
  -         '/transformed/b/source-2024-01-01-full-transformed-records-to-index.json',
  ?          -
  +         'transformed/b/source-2024-01-01-full-transformed-records-to-index.json',
        ],
    )
```

What I think this indicates is that the version as-is (that uses `os.walk()`) was including the leading slash `/`, thus making it an *absolute* path if used by anything downstream.  

While we certainly could just strip that, it feels like the use of `os.path.relpath` could be helpful in that it enforces a true relative path for a file, and not quite as much path string manipulation.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+                f"source: {container.labels['source']} | "
+                f"input_file: {container.labels['input_file']}\n"
+            )
+            file.write(header)
+            file.write(container.logs().decode())
+            file.write("\n\n")
+    logger.info(f"Log file created: {log_file}")
+
+    # update run.json with list of input files
+    update_or_create_run_json(run_directory, new_data={"input_files": input_files})
```
ghukill, line 81: Sorry, late comment add here...

Can we also write the transformed files to the `run.json`?  Maybe something like:

```json
"transformed_files": [
  [
    "transformed/a/alma-2024-08-29-daily-transformed-records-to-index.json"
  ],
  [
    "transformed/b/alma-2024-08-29-daily-transformed-records-to-index.json"
  ]
]
```

abdiff/core/run_ab_transforms.py
ghukill: Noting a quasi-related comment below: https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6#discussion_r1755319275

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Yes, it will -- once I update `containers=containers` -> `containers=running_containers` 😅 . Change has been made.

ghukill: Hahaha, I didn't even catch that!  Your change is how I had read the code.  Nice catch (agree that we only want to print out the timed-out containers).

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Good idea! Change has been made.

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Just to confirm: 

* If given a path that begins with a forward slash, `os.path.abspath` interprets the input as an absolute path. 
* If given a path that **does not** begin with a forward slash, `os.path.abspath` interprets the input as a relative path and returns a "normalized" path that **joins the current working directory with the input path**. 

Therefore, if we use `os.path.abspath`, we need to make sure that our current working directory is set to the base directory of the `transmogrifier-ab-diff` whenever we run `run_ab_transforms`. Is this how you understand it as well? 

Just wanted to make sure but totally on board to use `os.path.abspath` as needed!

ghukill: > Therefore, if we use os.path.abspath, we need to make sure that our current working directory is set to the base directory of the transmogrifier-ab-diff whenever we run run_ab_transforms. Is this how you understand it as well?

Kind of?  I think that 99% of our CLI apps assume that you are invoking them from the root of the project.  All of our relative imports are setup this way.  So I think it's pretty much implied this is what would happen.  

Related, no matter where you are, if you pass a relative path for something it must be relative to _where you are_.  So even then, `os.path.abspath()` would still work, as it'd be the (assumed correct) relative path + wherever you are.

To be clear: the use of `os.path.relpath()` or `os.path.abspath()` should **not** require ever doing any kind of `sys.path` setting or changing the working directory.  The real benefit of using `os.path.abspath()` would be: **if a relative path is used when identifying the job or run directory, it would be converted into an absolute path before mounting that into the Docker containers**.

Happy to talk more though!

abdiff/core/run_ab_transforms.py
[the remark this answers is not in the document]
jonavellecuerdo: Thank you, both! I have a better understanding of how `glob.glob` works and will remember to use `os.path.relpath` and `os.path.abspath` in the future. I updated `get_transformed_files` to follow the second suggestion from @ghukill , which uses `os.path.relpath` + `glob`.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+    )
+
+    # create list of run configs
+    run_configs = [
+        (image_name_a, transformed_directory_a),
+        (image_name_b, transformed_directory_b),
+    ]
+
+    # transform input files using images A and B via Docker containers
+    # save transformed record JSON files in corresponding directory
```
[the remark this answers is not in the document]
jonavellecuerdo: Change has been made -- removed in-line comments since docstring has been updated.

abdiff/core/utils.py
jonavellecuerdo: Docstring has been added.

tests/conftest.py
jonavellecuerdo: Change has been made.

abdiff/core/run_ab_transforms.py
```diff
@@ -1 +1,187 @@
...
+
+
+def run_docker_container(
+    docker_image: str,
+    transformed_directory: str,
+    input_file: str,
+    docker_client: docker.client.DockerClient,
+) -> Container:
+    """Run transmogrifier via Docker container to transform input file."""
+    source, output_filename = parse_transform_details_from_extract_filename(input_file)
```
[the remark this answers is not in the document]
jonavellecuerdo: Change has been made.

abdiff/core/run_ab_transforms.py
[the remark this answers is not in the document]
jonavellecuerdo: Change has been made, updated to: 

```python
log_file = Path(run_directory) / "transformed/logs.txt"
with open(log_file, "w") as file:
   ...
```

abdiff/core/run_ab_transforms.py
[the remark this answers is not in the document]
jonavellecuerdo: See [updated `get_transformed_files`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR204-R211).

abdiff/core/run_ab_transforms.py
jonavellecuerdo: See [updated docstring for `run_ab_transforms`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR32-R71).

tests/conftest.py
jonavellecuerdo: See [updated `MockedContainerRun.yield_mocked_run`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R49).

abdiff/core/run_ab_transforms.py
jonavellecuerdo: See new function [`aggregate_logs`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR178-R191).

abdiff/core/run_ab_transforms.py
jonavellecuerdo: See response in https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6#discussion_r1763591741.

abdiff/core/run_ab_transforms.py
jonavellecuerdo: See [updated lines of code for updating `run.json`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR112-R113).

jonavellecuerdo:
@ghukill @ehanson8 Please review the latest commit, addressing the most recent review comments. 😄 I resolved the comments to make it easier to navigate this PR but not before adding a response + link to relevant lines of code for each review comment. If there are any outstanding issues, feel free to un-resolve the comment to continue discussion. Thank you again for your helpful feedback!

abdiff/core/run_ab_transforms.py
```diff
@@ -28,13 +29,51 @@ def run_ab_transforms(
...
+
+    The following steps are perfomed in sequential order:
+        1. Directories are created to capture the transformed files.
+        2. For all input files, an A and B version of Transmogrifier is run
+           via detached Docker containers, resulting in all containers
+           running in parallel.
+        3. Wait for all containers to complete.
+        4. Aggregate logs from all containers.
+        5. Update run.json with lists describing input and transformed files.
+        5. Return a tuple containing two lists representing all A and B transformed files.
```
ghukill, line 42: Aces!  Thanks for adding.

abdiff/core/run_ab_transforms.py
```diff
@@ -43,63 +82,57 @@ def run_ab_transforms(
...
-        for container in containers:
-            header = (
-                f"docker_image: {container.labels['docker_image']} | "
-                f"source: {container.labels['source']} | "
-                f"input_file: {container.labels['input_file']}\n"
-            )
-            file.write(header)
-            file.write(container.logs().decode())
-            file.write("\n\n")
+    log_file = aggregate_logs(run_directory, containers)
```
ghukill, line 107: Commenting here arbitrarily, but thanks for the updates around here.  Looking at the full code locally, it's easy to scan and understand.

abdiff/core/run_ab_transforms.py
```diff
@@ -113,7 +146,7 @@ def run_docker_container(
             "source": source,
             "input_file": input_file,
         },
-        volumes=[f"{transformed_directory}:/tmp"],
+        volumes=[f"{os.path.abspath(transformed_directory)}:/tmp"],
```
ghukill, line 149: 👍 to `os.path.abspath()` here.  I'm unsure why docker containers need an absolute path (sure there is very good reason), but this definitely should satisfy it.

abdiff/core/run_ab_transforms.py
```diff
@@ -142,6 +175,22 @@ def wait_for_containers(containers: list[Container], timeout: int = 180) -> None
...
+        for container in containers:
+            header = (
+                f"docker_image: {container.labels['docker_image']} | "
+                f"source: {container.labels['source']} | "
+                f"input_file: {container.labels['input_file']}\n"
+            )
+            file.write(header)
+            file.write(container.logs().decode())
+            file.write("\n\n")
+    return log_file
```
ghukill, line 191: I like that it returns the log filepath; this feels consistent with files as input/ouput for most functionality in this application.

tests/test_run_ab_transforms.py
```diff
@@ -125,18 +142,18 @@ def test_get_transformed_files_success(
 
     assert get_transformed_files(run_directory) == (
         [
-            "/transformed/a/source-2024-01-01-full-transformed-records-to-index.json",
+            "transformed/a/source-2024-01-01-full-transformed-records-to-index.json",
```
ghukill, line 145: 🔥

ghukill (approved):
Looks good to me!!  Approved.

Thanks for all the back-and-forth and discussion @jonavellecuerdo.  Code reads great, feels like it covers the kind of usage patterns we expect.

Nice work!

abdiff/core/run_ab_transforms.py
```diff
@@ -43,63 +82,57 @@ def run_ab_transforms(
...
+
+    run_data = {"input_files": input_files, "transformed_files": transformed_files}
+    update_or_create_run_json(run_directory, run_data)
 
     elapsed_time = perf_counter() - start_time
     logger.info(
         "Total time to complete process: %s", str(timedelta(seconds=elapsed_time))
     )
-    return get_transformed_files(run_directory)
+    return transformed_files
```
ehanson8, line 119: Your updates have made this already great function even better!

abdiff/core/run_ab_transforms.py
```diff
@@ -142,6 +175,22 @@ def wait_for_containers(containers: list[Container], timeout: int = 180) -> None
         time.sleep(0.25)
 
 
+def aggregate_logs(run_directory: str, containers: list[Container]) -> str:
+    """Retrieve logs for containers in a list, aggregating to a single log file."""
+    log_file = str(Path(run_directory) / "transformed/logs.txt")
```
ehanson8, line 180: Looks like this `Path` call was missed when you removed the others

abdiff/core/utils.py
```diff
@@ -53,6 +53,14 @@ def update_or_create_run_json(run_directory: str, new_data: dict) -> dict:
...
+
+    This util is preferred for commands that require creating
+    nested subdirectories to organize outputs (e.g., run_ab_transforms).
+
+    Returns:
+        tuple[str, ...]: A tuple of absolute paths to created subdirectories.
+    """
     directories = []
     for subdirectory in subdirectories:
         directory = Path(base_directory) / subdirectory
```
ehanson8, line 66: Same with this one, it gets converted to a `str` anyway on line 68

ehanson8 (commented):
Great work! Just 2 small fixes

abdiff/core/run_ab_transforms.py
ghukill: (Extending / referencing the discussion from slack)

I think `Path` here _does_ make sense, as we are **constructing** a new Path from components.

abdiff/core/utils.py
jonavellecuerdo: See https://github.com/MITLibraries/transmogrifier-ab-diff/pull/6#discussion_r1765349539.

ehanson8 (approved):

PR #9, opened 2024-09-03
TIMX-340-build-ab-images

ehanson8:
Describe the overall purpose of the PR changes and any useful background context.

Start an `ipython` shell

```
pipenv run ipython
```

Run the following code to create Docker images for the 2 commits:

```
import docker
from abdiff.core.build_ab_images import build_ab_images

job_directory = "output/test-job"
commit_a = "008e20c"
commit_b = "395e612"
docker_client = docker.from_env()
build_ab_images(job_directory, docker_client, commit_a, commit_b)
```

While the code runs, you will see a temporary clone directory created for each commit which will be deleted after the Docker image is created.

After the code finishes running, you will see 2 Docker images created on your machine:

```
test-job-slug-008e20c:latest
test-job-slug-395e612:latest
```

You will also see `job.json` in `output/test-job` containing the same image names.

Delete the directory and Docker images after testing is complete.

YES

NO

- https://mitlibraries.atlassian.net/browse/TIMX-340

### Developer
- [x] All new ENV is documented in README
- [x] All new ENV has been added to staging and production environments
- [x] All related Jira tickets are linked in commit message(s)
- [x] Stakeholder approval has been confirmed (or is not needed)

### Code Reviewer(s)

### DIFF
--- a/abdiff/core/build_ab_images.py
+++ b/abdiff/core/build_ab_images.py
@@ -28,34 +28,39 @@ def build_ab_images(
     """
     if not docker_client:
         docker_client = docker.from_env()
-    image_names = []
+
+    image_tags = []
     for commit_sha in [commit_sha_a, commit_sha_b]:
         logger.debug(f"Processing commit: {commit_sha}")
-        image_name = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}:latest"
-        if docker_image_exists(docker_client, image_name):
-            image_names.append(image_name)
+        image_tag = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}:latest"
+        if docker_image_exists(docker_client, image_tag):
+            logger.debug("Docker image already exists with tag: {image_tag}")
+            image_tags.append(image_tag)
         else:
-            image = clone_repo_and_build_image(job_directory, commit_sha, docker_client)
-            image_names.append(image.tags[0])
+            image = build_image(job_directory, commit_sha, docker_client)
+            image_tags.append(image.tags[0])
         logger.debug(f"Finished processing commit: {commit_sha}")
-    images_data = {"image_name_a": image_names[0], "image_name_b": image_names[1]}
+
+    images_data = {"image_tag_a": image_tags[0], "image_tag_b": image_tags[1]}
     update_or_create_job_json(job_directory, images_data)
-    return (image_names[0], image_names[1])
+    return (image_tags[0], image_tags[1])
 
 
 def docker_image_exists(
-    docker_client: docker.client.DockerClient, image_name: str
+    docker_client: docker.client.DockerClient, image_tag: str
 ) -> bool:
     """Check if Docker image already exists with a certain name.
 
     Args:
         docker_client: A configured Docker client.
-        image_name: The name of the Docker image to be created.
+        image_tag: The tag of the Docker image to be created.
     """
-    return image_name in docker_client.images.list()
+    return image_tag in [
+        image_tag for image in docker_client.images.list() for image_tag in image.tags
+    ]
 
 
-def clone_repo_and_build_image(
+def build_image(
     job_directory: str,
     commit_sha: str,
     docker_client: docker.client.DockerClient,
@@ -68,9 +73,9 @@ def clone_repo_and_build_image(
         docker_client: A configured Docker client.
     """
     with tempfile.TemporaryDirectory() as clone_directory:
-        image_name = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}"
+        image_tag = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}"
         clone_repo_and_reset_to_commit(clone_directory, commit_sha)
-        image, _ = docker_client.images.build(path=clone_directory, tag=image_name)
+        image, _ = docker_client.images.build(path=clone_directory, tag=image_tag)
         logger.debug(f"Docker image created with tag: {image}")
         return image
 
--- a/abdiff/core/build_ab_images.py
+++ b/abdiff/core/build_ab_images.py
@@ -1,8 +1,9 @@
 import logging
-import os
-import shutil
+import tempfile
 
 import docker
+import docker.models
+import docker.models.images
 from pygit2 import clone_repository
 from pygit2.enums import ResetMode
 
@@ -13,29 +14,65 @@ logger = logging.getLogger(__name__)
 
 def build_ab_images(
     job_directory: str,
-    docker_client: docker.client.DockerClient,
     commit_sha_a: str,
     commit_sha_b: str,
+    docker_client: docker.client.DockerClient | None = None,
 ) -> tuple[str, str]:
     """Build Docker images based on 2 commit SHAs.
 
     Args:
         job_directory: The directory containing all files related to a job.
-        docker_client: A configured Docker client.
         commit_sha_a: The SHA of the first commit for comparison.
         commit_sha_b: The SHA of the second commit for comparison.
+        docker_client: A configured Docker client.
     """
-    images = []
+    if not docker_client:
+        docker_client = docker.from_env()
+    image_names = []
     for commit_sha in [commit_sha_a, commit_sha_b]:
-        clone_directory = f"{job_directory}/{commit_sha}"
-        image_name = f"{job_directory.split("/")[-1]}-{commit_sha}"
+        logger.debug(f"Processing commit: {commit_sha}")
+        image_name = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}:latest"
+        if docker_image_exists(docker_client, image_name):
+            image_names.append(image_name)
+        else:
+            image = clone_repo_and_build_image(job_directory, commit_sha, docker_client)
+            image_names.append(image.tags[0])
+        logger.debug(f"Finished processing commit: {commit_sha}")
+    images_data = {"image_name_a": image_names[0], "image_name_b": image_names[1]}
+    update_or_create_job_json(job_directory, images_data)
+    return (image_names[0], image_names[1])
+
+
+def docker_image_exists(
+    docker_client: docker.client.DockerClient, image_name: str
+) -> bool:
+    """Check if Docker image already exists with a certain name.
+
+    Args:
+        docker_client: A configured Docker client.
+        image_name: The name of the Docker image to be created.
+    """
+    return image_name in docker_client.images.list()
+
+
+def clone_repo_and_build_image(
+    job_directory: str,
+    commit_sha: str,
+    docker_client: docker.client.DockerClient,
+) -> docker.models.images.Image:
+    """Clone repo and build Docker image.
+
+    Args:
+        job_directory: The directory containing all files related to a job.
+        commit_sha: The SHA of the commit.
+        docker_client: A configured Docker client.
+    """
+    with tempfile.TemporaryDirectory() as clone_directory:
+        image_name = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}"
         clone_repo_and_reset_to_commit(clone_directory, commit_sha)
         image, _ = docker_client.images.build(path=clone_directory, tag=image_name)
-        images.append(image.tags[0])
-        delete_repo_clone(clone_directory)
-    images_update = {"image_name_a": images[0], "image_name_b": images[1]}
-    update_or_create_job_json(job_directory, images_update)
-    return (images[0], images[1])
+        logger.debug(f"Docker image created with tag: {image}")
+        return image
 
 
 def clone_repo_and_reset_to_commit(clone_directory: str, commit_sha: str) -> None:
@@ -53,15 +90,3 @@ def clone_repo_and_reset_to_commit(clone_directory: str, commit_sha: str) -> Non
     logger.debug(f"Cloned repo to: {clone_directory}")
     repository.reset(commit_sha, ResetMode.HARD)
     logger.debug(f"Cloned repo reset to commit: {commit_sha}")
-
-
-def delete_repo_clone(clone_directory: str) -> None:
-    """Remove cloned repository.
-
-    Args:
-        clone_directory: The directory for the cloned repo.
-    """
-    logger.debug(f"Removing repo clone: {clone_directory}")
-    if os.path.exists(clone_directory):
-        shutil.rmtree(clone_directory)
-    logger.debug(f"Removed repo clone: {clone_directory}")
--- a/abdiff/core/build_ab_images.py
+++ b/abdiff/core/build_ab_images.py
@@ -34,7 +34,7 @@ def build_ab_images(
         logger.debug(f"Processing commit: {commit_sha}")
         image_tag = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}:latest"
         if docker_image_exists(docker_client, image_tag):
-            logger.debug("Docker image already exists with tag: {image_tag}")
+            logger.debug(f"Docker image already exists with tag: {image_tag}")
             image_tags.append(image_tag)
         else:
             image = build_image(job_directory, commit_sha, docker_client)
--- a/abdiff/core/build_ab_images.py
+++ b/abdiff/core/build_ab_images.py
@@ -1,1 +1,67 @@
-"""abdiff.core.build_ab_images"""
+import logging
+import os
+import shutil
+
+import docker
+from pygit2 import clone_repository
+from pygit2.enums import ResetMode
+
+from abdiff.core.utils import update_or_create_job_json
+
+logger = logging.getLogger(__name__)
+
+
+def build_ab_images(
+    job_directory: str,
+    docker_client: docker.client.DockerClient,
+    commit_sha_a: str,
+    commit_sha_b: str,
+) -> tuple[str, str]:
+    """Build Docker images based on 2 commit SHAs.
+
+    Args:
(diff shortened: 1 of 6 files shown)


abdiff/core/build_ab_images.py
```diff
@@ -1 +1,67 @@
...
+from pygit2.enums import ResetMode
+
+from abdiff.core.utils import update_or_create_job_json
+
+logger = logging.getLogger(__name__)
+
+
+def build_ab_images(
+    job_directory: str,
+    docker_client: docker.client.DockerClient,
```
ehanson8, line 16: This is a shift from what we've previously discussed but I figure the same docker client would be used at here and in the subsequent `run_ab_transforms` so the Docker client could be configured once at the CLI-level and passed to each. But let me know if this is a bad idea!

tests/test_build_ab_images.py
```diff
@@ -0,0 +1,43 @@
...
+    assert images[0] == "987abc"
+    assert images[1] == "654def"
+    assert read_job_json(job_directory) == {
+        "image_name_a": "987abc",
+        "image_name_b": "654def",
+    }
+
+
+@patch("abdiff.core.build_ab_images.clone_repository")
+def test_clone_repo_from_commit_success(mocked_clone, job_directory):
```
ehanson8, line 27: Oops, need to rename this to the new function name!

abdiff/core/build_ab_images.py
jonavellecuerdo: This is a question that's come up whenever I write code that involves `boto3` clients, so thank you for bringing this up and looking forward to discussing as a team. I came across [this article](https://www.digitalocean.com/community/questions/should-i-create-a-new-instance-of-boto3-client-for-each-file-upload-request-or-use-a-shared-instance), which recommends **reusing** client instances as "each new client instance consumes resources (e.g., memory, file descriptors for connections)" and "creating a new client for every request can add significant overhead". I also tried asking this question to ChatGPT, and this was the response:
> Creating a new DockerClient instance involves establishing a new connection to the Docker daemon, which can be resource-intensive. By reusing a single instance, you minimize the overhead associated with repeatedly opening and closing connections.

I am leaning towards reusing clients (as that is what I'm also seeing in various StackOverflow threads / articles), but interested in what @ghukill 's thinks as well!

ghukill: Interesting, yeah, glad we're discussing!

My knee-jerk is that I'm a little wary having the CLI be responsible for providing a docker client, which makes this function somewhat dependent on the CLI context.  

That said... becuase you have it as a typed argument, it's _very_ clear what it needs to run.

What about a bit of a hybrid:
1. keep the argument, but make it optional/positional with a default of `None`
2. if `None` within the method call, then instantiate then via `docker_client = docker.from_env()`

To me that advantages are:
- from the CLI context, you can initialize it and pass to any functions that may need it, knowing you've prevented them from creating their own
- this function still works simply without the the CLI context

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,67 @@
...
+
+    Args:
+        job_directory: The directory containing all files related to a job.
+        docker_client: A configured Docker client.
+        commit_sha_a: The SHA of the first commit for comparison.
+        commit_sha_b: The SHA of the second commit for comparison.
+    """
+    images = []
+    for commit_sha in [commit_sha_a, commit_sha_b]:
+        clone_directory = f"{job_directory}/{commit_sha}"
```
ghukill, line 30: If I'm understand this correctly, this is temporarily cloning transmogrifier into the job directory under a particular commit SHA?

I wonder if this is a good use for [python's `tempfile` module](https://docs.python.org/3/library/tempfile.html)?  This provides a temporary location on disk to do things, but is _automatically_ cleaned up after the python process closes. Example of a getting a temp dir:

```python
with tempfile.TemporaryDirectory() as tmpdir:
    print(tmpdir)
# /var/folders/26/b99kw7554l7bgrx51srd_7tr0000gp/T/tmpr7gdefsg
```

In this way, it's possible you may not need the method `delete_repo_clone()` at all, as we could rely on `tempfile` to automatically cleanup when the python process completes.

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,67 @@
...
+    """Build Docker images based on 2 commit SHAs.
+
+    Args:
+        job_directory: The directory containing all files related to a job.
+        docker_client: A configured Docker client.
+        commit_sha_a: The SHA of the first commit for comparison.
+        commit_sha_b: The SHA of the second commit for comparison.
+    """
+    images = []
+    for commit_sha in [commit_sha_a, commit_sha_b]:
```
ghukill, line 29: What if we created a helper function to perform this work?  Then this parent/driver function is just responsible for organizing arguments and preparing things, but a little, terse function actually performs the clone + build?

Seems like it could be easier to test, but I haven't looked at the tests yet in detail.

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,67 @@
...
+        clone_directory = f"{job_directory}/{commit_sha}"
+        image_name = f"{job_directory.split("/")[-1]}-{commit_sha}"
```
ghukill, line 31: If a job directory is `/my/path/to/super-job`, would the image name would then be `super-job-abc123`?

Would it be worth pre-pending the image name with `transmogrifier` so we know that's what the image is?  Making it: `transmogrifier-super-job-abc123`.

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,67 @@
...
+import docker
+from pygit2 import clone_repository
+from pygit2.enums import ResetMode
+
+from abdiff.core.utils import update_or_create_job_json
+
+logger = logging.getLogger(__name__)
+
+
+def build_ab_images(
```
ghukill, line 14: Leaving this comment here at the function level, given other comments in the logic itself.

Would it be worth seeing if an image is already built for a particular SHA?  In a comment below, noted that an image name could be `super-job-abc123` (or `transmogrifier-super-job-abc123` if we include that), and we'd know this as a combination of the job name + SHA.

As such, wonder if we could look for a docker image _already built_ with this image name.  If so, then it seems like we could skip (re)building the image and just be confident it exists, but still include in the `job.json` and return as normal (as if it had been created).

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,67 @@
...
+    repository = clone_repository(
+        "https://github.com/MITLibraries/transmogrifier.git",
+        clone_directory,
+    )
+    logger.debug(f"Cloned repo to: {clone_directory}")
+    repository.reset(commit_sha, ResetMode.HARD)
+    logger.debug(f"Cloned repo reset to commit: {commit_sha}")
+
+
+def delete_repo_clone(clone_directory: str) -> None:
```
ghukill, line 58: As noted above about possible use of `tempfile`, maybe not needed if going that route.

ghukill (commented):
Looking good!  For the record, running the manual tests worked flawlessly: I had two images built, from the commit SHAs I provided.  Very cool.

I left a handful of comments in the terse, main `build_ab_images` function.  

My primary questions boil down to:
1. is it worth checking if an image is already built, and if so, reuse?
2. what about using `tempfile` to provide an automatically cleaned up location on disk for cloning transmog
3. consider adding `transmogrifier-` as a prefix to image names to disambiguate them among other docker images

I _also_ wonder about overridding the git clone + commit reset, supposing I had a version of Transmog locally I want to use... but I'd vote to create a ticket for this and consider at a later time.  Keep it simple :tm: for now.

Overall, nice work!  Most of my comments are ergonomics and/or optional, but the functionality appears to be there.

abdiff/core/build_ab_images.py
ghukill: Depending on if, and how, this is refactored to a standlone function, I do think some logging could be helpful as well confirming that bits and pieces have been performed.  But highly dependent on if/how any changes are made during this PR.

abdiff/core/build_ab_images.py
jonavellecuerdo: I agree with pre-pending `transmogrifier` to the image name.

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,67 @@
...
+    """
+    images = []
+    for commit_sha in [commit_sha_a, commit_sha_b]:
+        clone_directory = f"{job_directory}/{commit_sha}"
+        image_name = f"{job_directory.split("/")[-1]}-{commit_sha}"
+        clone_repo_and_reset_to_commit(clone_directory, commit_sha)
+        image, _ = docker_client.images.build(path=clone_directory, tag=image_name)
+        images.append(image.tags[0])
+        delete_repo_clone(clone_directory)
+    images_update = {"image_name_a": images[0], "image_name_b": images[1]}
```
jonavellecuerdo, line 36: Minor change request: Can we rename `images_updates` -> `images_data` to align with the naming convention for the parameter passed into `update_or_create_job_json` used for `init_job` and `init_run`? 🤔

abdiff/core/build_ab_images.py
jonavellecuerdo: I don't have a strong preference myself, but I would like to keep the `images_update` and `update_or_create_job_json` _outside_ of the helper function such that -- the helper function is specific to _this command_ while the `update_or_create_job_json` is a universal step across all commands.

jonavellecuerdo (commented):
Hi @ehanson8 -- awesome first pass! Had a couple comments but generally agreed with @ghukill 's proposed suggestions.

abdiff/core/build_ab_images.py
ghukill: If this helper function were created -- and big "if", unsure if needed -- I'd agree with @jonavellecuerdo that updating the job JSON would be outside its scope.  That feels like the work of the "core" function.

abdiff/core/build_ab_images.py
ehanson8: That works for me, best of both worlds!

abdiff/core/build_ab_images.py
ehanson8: This is great and was simple to implement, great suggestion!

abdiff/core/build_ab_images.py
ehanson8: Done!

abdiff/core/build_ab_images.py
ehanson8: Correct, it has been removed!

abdiff/core/build_ab_images.py
ehanson8: Good call!

abdiff/core/build_ab_images.py
ehanson8: Created a helper function, it does make the code more readable

ghukill:
@ehanson8 - one request that I forgot during my first review: can we add a `build_ab_images` to the list of importable functions for `abdiff.core` like is done here?

https://github.com/MITLibraries/transmogrifier-ab-diff/blob/main/abdiff/core/__init__.py#L6-L9

ehanson8:
> @ehanson8 - one request that I forgot during my first review: can we add a `build_ab_images` to the list of importable functions for `abdiff.core` like is done here?
> 
> https://github.com/MITLibraries/transmogrifier-ab-diff/blob/main/abdiff/core/__init__.py#L6-L9

Will do!

abdiff/core/build_ab_images.py
ehanson8: Great idea, added a function to do this!

ehanson8:
@ghukill @jonavellecuerdo Pushed a new commit, great suggestions!

ghukill (approved):
Approved!  I left a couple of stylstic comments/suggestions, but comfortable approving with or without the changes.

Nice work!  Lot happening here: git cloning, docker building, but the code feels understandable and succinct.

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,92 @@
...
+def docker_image_exists(
+    docker_client: docker.client.DockerClient, image_name: str
+) -> bool:
+    """Check if Docker image already exists with a certain name.
+
+    Args:
+        docker_client: A configured Docker client.
+        image_name: The name of the Docker image to be created.
+    """
+    return image_name in docker_client.images.list()
```
ghukill, line 55: Update: after some local testing, I'm not sure this `image_name in docker_client.images.list()` works as expected.

The output I'm seeing from `.images.list()` looks like they are `Image` instances:
```
[<Image: 'transmogrifier-hello-world-1910426fbcc98e098de222ccdcce3cf3ac1337e7:latest'>,
 <Image: 'transmogrifier-hello-world-395e612f8887b279ccf35363694bafb06ef9a25a:latest'>,
 <Image: ''>,
 <Image: ''>,
 <Image: 'test-job-395e612:latest'>,
 <Image: 'test-job-008e20c:latest'>,
...
...
]
```

But we're passing a string, so it seems possible the `in` operator doesn't find a string in a list of `Image` instances.

Wonder if this needs to not only get the images, but also then grab the first (or all?) image tags? 

Example:
```python
image = docker_client.images.list()[0]

image.tags[0]
# Out[13]: 'transmogrifier-hello-world-1910426fbcc98e098de222ccdcce3cf3ac1337e7:latest'
# this look like what we'd want to match on...
```

ghukill (changes requested):
Everything from previous review stands, but am requesting a change here for the `docker_image_exists()` function.  Looks like maybe it needs a tweak for the string matching to work?

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,92 @@
...
+        image_name = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}:latest"
+        if docker_image_exists(docker_client, image_name):
+            image_names.append(image_name)
+        else:
+            image = clone_repo_and_build_image(job_directory, commit_sha, docker_client)
+            image_names.append(image.tags[0])
+        logger.debug(f"Finished processing commit: {commit_sha}")
+    images_data = {"image_name_a": image_names[0], "image_name_b": image_names[1]}
+    update_or_create_job_json(job_directory, images_data)
+    return (image_names[0], image_names[1])
```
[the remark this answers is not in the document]
ehanson8: Good call! I don't think readability is a nitpicky concern at all

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,92 @@
...
+    """Check if Docker image already exists with a certain name.
+
+    Args:
+        docker_client: A configured Docker client.
+        image_name: The name of the Docker image to be created.
+    """
+    return image_name in docker_client.images.list()
+
+
+def clone_repo_and_build_image(
```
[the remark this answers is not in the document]
ehanson8: Good point, changed!

tests/test_build_ab_images.py
```diff
@@ -0,0 +1,70 @@
...
+    build_ab_images,
+    clone_repo_and_build_image,
+    clone_repo_and_reset_to_commit,
+    docker_image_exists,
+)
+from abdiff.core.utils import read_job_json
+
+
+@patch("abdiff.core.build_ab_images.clone_repository")
+def test_build_ab_images_success(mocked_clone, job_directory, mocked_docker_client):
```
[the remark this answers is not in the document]
ehanson8: First used this to mock `oracledb` in `alma-patronload` so the app didn't accidentally open a connection to the DW during testing

abdiff/core/build_ab_images.py
ehanson8: I did manually test this but must have botched it, you are correct and I've refactored!

ehanson8:
@ghukill @jonavellecuerdo Another commit of fixes!

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,97 @@
...
+) -> bool:
+    """Check if Docker image already exists with a certain name.
+
+    Args:
+        docker_client: A configured Docker client.
+        image_tag: The tag of the Docker image to be created.
+    """
+    return image_tag in [
+        image_tag for image in docker_client.images.list() for image_tag in image.tags
+    ]
```
ghukill, line 60: The ole nested list comprehension!  🔥 .

Tested on a rebuild, works great!  Skipped a rebuild.

abdiff/core/build_ab_images.py
```diff
@@ -1 +1,97 @@
...
+    """
+    if not docker_client:
+        docker_client = docker.from_env()
+
+    image_tags = []
+    for commit_sha in [commit_sha_a, commit_sha_b]:
+        logger.debug(f"Processing commit: {commit_sha}")
+        image_tag = f"transmogrifier-{job_directory.split("/")[-1]}-{commit_sha}:latest"
+        if docker_image_exists(docker_client, image_tag):
+            logger.debug("Docker image already exists with tag: {image_tag}")
```
ghukill, line 37: Sorry to prompt another commit, but believe this needs an `f`-string prefix, just a normal string as-is.  This was the output in my local testing:

```
Docker image already exists with tag: {image_tag}
```

ghukill (approved):
Thanks for the updates.  Other than one very minor requested change (missing f-string), all looks good to me.  But please consider this approved from my POV.

abdiff/core/build_ab_images.py
ehanson8: Ugh, good catch!

jonavellecuerdo (approved):

PR #22, opened 2024-10-04
Add 'collate_ab_transforms' command

jonavellecuerdo:
Implement function to combine transformed records (in JSON files) generated by A/B Docker images of `transmogrifier` into a single [Parquet dataset](https://arrow.apache.org/docs/python/dataset.html).

I focused on providing more detail in the function docstrings, so I highly recommend giving these a read during your review as it will help your understanding of this workflow. The [docstring for `collate_ab_transforms`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/22/files#diff-1cf5419c409425ac3b8275acb901b7c68cc1131a5b24d5846f477eb967681e55R40-R50) provides a good summary, that I'll reiterate shortly here: 

* There are two (2) parquet datasets at work here:
   * The **first** is a _temporary_, intermediate Parquet dataset that contains TIMDEX record dictionaries (i.e., the JSON file of transformed records converted to Parquet). See [schema](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/22/files#diff-1cf5419c409425ac3b8275acb901b7c68cc1131a5b24d5846f477eb967681e55R18-R26).
   * The **second** is a _persistent_ Parquet dataset that contains the joined records, which are generated by querying the _first_ dataset using DuckDB to join A/B versions of transformed records.  See [schema](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/22/files#diff-1cf5419c409425ac3b8275acb901b7c68cc1131a5b24d5846f477eb967681e55R27-R34). 
      * **NOTE:** I note that this dataset is "persistent", but @ghukill has noted potentially removing this dataset later in the workflow. Noting for awareness but to be discussed during development of the downstream functions (i.e., `calc_ab_diffs`).

Running `make test` and confirming that all corresponding unit tests pass should be sufficient for reviewing these changes. Here are a couple items worth noting: 

* I copied over the `transformed` directory created by my previous testing of `run_ab_transforms` as test fixtures. Previous testing resulted in this transformed file: `alma-2024-08-29-daily-transformed-records-to-index.json`. [I describe how I manually edited the files for purposes of testing of _this_ function in the docstring for `test_get_joined_batches_iter_success`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/22/files#diff-487a91147988fd33daa9cc446cf6aa3f819b1f2cdc01bff8184e2f047ac808c5R64-R73).
* When `test_collate_ab_transforms_success` is run, it creates the `collated` Parquet dataset in the fixtures directory. I ran this test and committed the created dataset.

[YES](https://mitlibraries.atlassian.net/browse/TIMX-349); added `duckdb`, `ijson`, and `pyarrow` as dependencies.

NO

- https://mitlibraries.atlassian.net/browse/TIMX-349

### Developer

### Code Reviewer(s)


abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+        f"Wrote {len(transformed_written_files)} parquet file(s) to transformed dataset"
+    )
+
+    joined_written_files = write_to_dataset(
+        get_joined_batches_iter(transformed_dataset_path.name),
+        base_dir=collated_dataset_path,
+        schema=JOINED_DATASET_SCHEMA,
+    )
+    logger.info(f"Wrote {len(joined_written_files)} parquet file(s) to collated dataset")
+    return collated_dataset_path
```
ehanson8, line 70: Very readable, great work!

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+    * record: The TIMDEX record serialized to a JSON string then encoded to bytes.
+    * version: The version of the transform, parsed from the absolute filepath to a
+      transformed file.
+    * transformed_file_name: The name of the transformed file, excluding file extension.
+    """
+    version, transformed_file_name = parse_parquet_details_from_transformed_file(
+        str(transformed_file)
+    )
+    with open(transformed_file) as file:
+        records = ijson.items(file, "item")
```
ghukill, line 92: I think that opening the file with mode `"rb"` will avoid a linting warning that I was seeing in the console:

```python
with open(transformed_file, "rb") as file:
```

Warning:
```
tests/test_collate_ab_transforms.py::test_get_joined_batches_iter_success
  ...python3.12/site-packages/ijson/compat.py:32: DeprecationWarning: 
  ijson works by reading bytes, but a string reader has been given instead. This
  probably, but not necessarily, means a file-like object has been opened in text
  mode ('t') rather than binary mode ('b').
  
  An automatic conversion is being performed on the fly to continue, but on the
  other hand this creates unnecessary encoding/decoding operations that decrease
  the efficiency of the system. In the future this automatic conversion will be
  removed, and users will receive errors instead of this warning. To avoid this
  problem make sure file-like objects are opened in binary mode instead of text
  mode.
  
    warnings.warn(_str_vs_bytes_warning, DeprecationWarning)
```

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+    joined_written_files = write_to_dataset(
+        get_joined_batches_iter(transformed_dataset_path.name),
+        base_dir=collated_dataset_path,
+        schema=JOINED_DATASET_SCHEMA,
+    )
+    logger.info(f"Wrote {len(joined_written_files)} parquet file(s) to collated dataset")
+    return collated_dataset_path
+
+
+def yield_records(transformed_file: str | Path) -> Generator:
```
ehanson8, line 73: I think the name is misleading, maybe `yield_data_from_records`? Otherwise, I assume full records as the output. Also, I would expand the type hint to `Generator[dict[str,str]` to further clarify the output

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+    This function uses ijson to yield records from a JSON stream
+    (i.e., the transformed file) one at a time. A generator is returned,
+    yielding a dictionary that contains the following:
+
+    * timdex_record_id: The TIMDEX record ID.
+    * source: The shorthand name of the source as denoted in by Transmogrifier
+      (see https://github.com/MITLibraries/transmogrifier/blob/main/transmogrifier/config.py).
+    * record: The TIMDEX record serialized to a JSON string then encoded to bytes.
+    * version: The version of the transform, parsed from the absolute filepath to a
+      transformed file.
```
ehanson8, line 85: Is `version` the same as commit SHA or something else?

tests/test_collate_ab_transforms.py
```diff
@@ -0,0 +1,108 @@
+# ruff: noqa: PLR2004
+
+from pathlib import Path
+
+import pyarrow as pa
+import pytest
+
```
ghukill, line 7: Noting this here, as a little unsure where to note otherwise.  

Can we add `collate_ab_transforms` to the `abdiff.core.__init__` file, so it's directly importable?  Similar to how we've done other core functions?

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+    * transformed_file_name: The name of the transformed file, excluding file extension.
+    """
+    version, transformed_file_name = parse_parquet_details_from_transformed_file(
+        str(transformed_file)
+    )
+    with open(transformed_file) as file:
+        records = ijson.items(file, "item")
+        for record in records:
```
ehanson8, line 93: I would just do `for record in ijson.items(file, "item")`, no need for the `records` variable

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+                "source": record["source"],
+                "record": json.dumps(record).encode(),
+                "version": version,
+                "transformed_file_name": transformed_file_name,
+            }
+
+
+def get_transformed_batches_iter(
+    run_directory: str, transformed_files: tuple[list[str], ...]
+) -> Generator:
```
ehanson8, line 105: Again, be specific about what the `Generator` is returning in the type hint

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+    TIMDEX records and compiling the dictionaries into a
+    pyarrow.RecordBatch. The size of a batch is set by the
+    'READ_BATCH_SIZE' global variable.
+
+    The function returns a generator, yielding batches of the dictionaries
+    from yield_records(). The returned generator can be passed to
+    abdiff.core.utils.write_to_dataset() to perform batch writes to
+    a Parquet dataset.
+    """
+    for transformed_files_list in transformed_files:
```
ehanson8, line 119: Optional: `for transformed_files_list in transformed_file_lists:`, a bit of whiplash from shifting variable names in the next few lines

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+        for transformed_file in transformed_files_list:
+            record_iter = yield_records(
+                transformed_file=Path(run_directory) / transformed_file
+            )
+            for record_batch in itertools.batched(record_iter, READ_BATCH_SIZE):
+                yield pa.RecordBatch.from_pylist(list(record_batch))
+
+
+def get_joined_batches_iter(dataset_directory: str) -> Generator:
```
ehanson8, line 128: More specific `Generator` type hint

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+    from yield_records(). The returned generator can be passed to
+    abdiff.core.utils.write_to_dataset() to perform batch writes to
+    a Parquet dataset.
+    """
+    for transformed_files_list in transformed_files:
+        for transformed_file in transformed_files_list:
+            record_iter = yield_records(
+                transformed_file=Path(run_directory) / transformed_file
+            )
+            for record_batch in itertools.batched(record_iter, READ_BATCH_SIZE):
```
ehanson8, line 124: Nice use of the new method!

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+    Args:
+        dataset_directory: The root directory of the Parquet dataset of TIMDEX records
+            (i.e., the tempfile.TemporaryDirectory).
+    """
+    with duckdb.connect() as con:
+        transformed_file_names = con.execute(
+            f"""
+            SELECT DISTINCT transformed_file_name FROM
+            read_parquet('{dataset_directory}/**/*.parquet', hive_partitioning=true)
+            """
```
ehanson8, line 170: These queries perhaps would be better as meaningfully-named variables, the function would be more readable that way in my opinion

ehanson8 (commented):
Looking very good, a lot of complicated work that you've made very readable!

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+        ).fetchall()[0]
```
ghukill, line 171: As discussed in a quick huddle, this `[0]` index call is problematic if there are **multiple** transform files as part of the run.

Had discussed a couple of options are:
1. list comprehension to pull out the transform file name from each row returned
2. install `pandas` and use `.fetchdf()[COLUMN_NAME]` to get a list of values for that column in the SQL query

Thankfully, quite minor!  With that change, confirmed that the `for transformed_file in transformed_file_names` block later in this method successfully yields batches from each transform file, resulting in an expected number of rows in the final dataset.

tests/test_collate_ab_transforms.py
```diff
@@ -0,0 +1,108 @@
...
+    transformed_batches_iter = get_transformed_batches_iter(
+        run_directory=example_run_directory, transformed_files=example_transformed_files
+    )
+    transformed_batch = next(transformed_batches_iter)
+
+    assert isinstance(transformed_batch, pa.RecordBatch)
+    assert transformed_batch.num_rows <= READ_BATCH_SIZE
+    assert transformed_batch.schema == TRANSFORMED_DATASET_SCHEMA
+
+
```
ghukill, line 62: Arbitrary location of comment: as discussed, probably a good idea to add a test that confirms if **multiple** transform files are provided (e.g. `alma` _and_ `dspace`), that the resulting dataset contains joined rows from both.

ghukill (changes requested):
Marking this as "Request changes" for the small bug in looping through transform files before joining and writing. 

That said... looks good to me!  Nice work all around.  Easy to follow what methods do, good docstrings.  I think they are flirting with overly descriptive versus pointing out the particular complex parts... but I'm good with them.  We'll just need to ensure they don't become inaccurate if things change.

Left a couple of suggestions -- e.g. parameters for DuckDB SQL queries -- but all just suggestions, all seemed functional.

Nice work again!

abdiff/core/collate_ab_transforms.py
ghukill: I'd second this - forgot to comment on my review.  Maybe the input argument of `transformed_file_lists` could more accurately be something like `a_and_b_transformed_file_lists` (though kind of a mouthful).

abdiff/core/collate_ab_transforms.py
ghukill: Flip side to that: if we use DuckDB SQL query parameters [as proposed here](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/22#discussion_r1788000257), then it complicates moving these to a standalone variable.

Or, actually, it probably wouldn't!  Mentioning that dimension, but I think it would work where the string just has `?` characters in it, and DuckDB connection knows to replace them with passed parameters when it's used.

Pipfile
```diff
@@ -6,6 +6,9 @@ name = "pypi"
 [packages]
 click = "*"
 docker = "*"
+duckdb = "*"
+ijson = "*"
+pyarrow = "*"
```
ghukill, line 11: Commenting here, but would be a `dev package.

Might be worth installing `pyarrow-stubs`?  In some local testing, I've found it helpful to expose to raise linting errors when types might be functional, but not ideal.

Example from `collate_ab_transforms.py`:
```python
TRANSFORMED_DATASET_SCHEMA = pa.schema(
    [
        pa.field("timdex_record_id", pa.string()),
        pa.field("source", pa.string()),
        pa.field("record", pa.binary()),
        pa.field("version", pa.string()),
        pa.field("transformed_file_name", pa.string()),
    ]
)
JOINED_DATASET_SCHEMA = pa.schema(
    [
        pa.field("timdex_record_id", pa.string()),
        pa.field("source", pa.string()),
        pa.field("record_a", pa.binary()),
        pa.field("record_b", pa.binary()),
    ]
)
```

As is, this will raise a linting error where `pa.schema` cannot accept a list.  But converting them to a tuple works!

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+
+import duckdb
+import ijson
+import pyarrow as pa
+
+from abdiff.core.utils import write_to_dataset
+
+logger = logging.getLogger(__name__)
+
+READ_BATCH_SIZE = 1_000
```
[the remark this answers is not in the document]
jonavellecuerdo: Ah, it is just [syntax to make it easier to read through long literals](https://peps.python.org/pep-0515/). :)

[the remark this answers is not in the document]
ehanson8: Oh, new to me, thanks!

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: Hmm, what do you think about renaming to `get_transformed_records_iter` (to be consistent with the other `Generator` functions in `collate_ab_transforms.py`)? Also curious to get @ghukill 's as he helped with this function. 🤔 

As a reminder, this is what the function _yields_:

```python
 yield {
   "timdex_record_id": record["timdex_record_id"],
   "source": record["source"],
   "record": json.dumps(record).encode(),
   "version": version,
   "transformed_file_name": transformed_file_name,
}
```
It returns the TIMDEX record _serialized to a JSON string_ + other vars to help with the dataset creation downstream...

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: Good idea! Made the change.

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: Indeed! Looking forward to using it when the need arises! 👀

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: Updated return type hint to `Generator[pa.RecordBatch]`.

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: Updated return type hint to `Generator[pa.RecordBatch]`.

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: Hmm, what do ya'll think of the following?:

```python

def collate_ab_transforms(run_directory: str, ab_transformed_file_lists: tuple[list[str], ...])
   ...

def get_transformed_batches_iter(run_directory: str, ab_transformed_file_lists: tuple[list[str], ...]) -> Generator[pa.RecordBatch]:
   for transformed_files in ab_transformed_file_lists:
      for transformed_file in transformed_files:
   ...
```

**Note:** Pretty much @ghukill 's suggestion without the conjunction "and" to keep it consistent with our other function names (using "ab").

abdiff/core/collate_ab_transforms.py
ehanson8: My apologies, I was completely misreading this so please ignore my objections!

ghukill: `get_transformed_records_iter()` makes sense to me!

I think `yield_data_from_records()` was kind of a stubbed / half-baked name, apologies.  I like this better.

abdiff/core/collate_ab_transforms.py
ghukill: I dig it!

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+                    b AS (SELECT * FROM transformed_file WHERE version='b')
+                SELECT
+                    a.timdex_record_id,
+                    a.source,
+                    a.record as record_a,
+                    b.record as record_b
+                FROM a
+                FULL OUTER JOIN b USING (timdex_record_id)
+                """
+            ).fetch_record_batch(READ_BATCH_SIZE)
```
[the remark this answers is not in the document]
jonavellecuerdo: Ah, thank you for bringing this up! Ruff does raise an error/warning for rule [S608](https://docs.astral.sh/ruff/rules/hardcoded-sql-expression/#hardcoded-sql-expression-s608) re: SQL injection. I updated `pyproject.toml` to ignore the rule, _but_ I do prefer to update the syntax over ignoring when possible (even when risk is low - i.e., this being an internal tool). I will update to query to [use the syntax for a prepared statement via "named parameters" ](https://duckdb.org/docs/api/python/dbapi.html#prepared-statements).

abdiff/core/collate_ab_transforms.py
ghukill: Jumping in here arbitrarily, but appreciate these more detailed type hints requests.  Don't _always_ think they have lots of value, but when we're interacting with a newish library to us, or one that has unusual inputs/outputs, I think it's quite helpful.

abdiff/core/collate_ab_transforms.py
ehanson8: Looks good!

abdiff/core/collate_ab_transforms.py
```diff
@@ -0,0 +1,220 @@
...
+
+            while True:
+                try:
+                    yield results.read_next_batch()
+                except StopIteration:
+                    logger.info("Already fetched all batches.")
```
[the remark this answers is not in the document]
jonavellecuerdo: Removed.

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: Hmm, I don't seem to be getting this warning? 🤔

tests/test_collate_ab_transforms.py
jonavellecuerdo: Updated.

abdiff/core/collate_ab_transforms.py
ghukill: @jonavellecuerdo - I have a hunch (but could be wrong) that it's an `ijson` version thing?  I wonder if I have a newer version which is showing this.  I'm basing this hunch on the warning which suggests in _future_ versions, it will actually throw an error instead of a warning.

My suggestion would be to add now, to get ahead of this warning/error, regardless of `ijson` version.

jonavellecuerdo: Updated to use file mode `"rb"`.

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: Updated!

tests/test_collate_ab_transforms.py
jonavellecuerdo: Updated. Please see latest commit https://github.com/MITLibraries/transmogrifier-ab-diff/pull/22/commits/c8c00da40bb24322d5d28ea4b932895c690d80de.

tests/test_collate_ab_transforms.py
```diff
@@ -12,21 +13,54 @@
...
     assert collated_dataset_path == str(Path(example_run_directory) / "collated")
 
+    collated_dataset = ds.dataset(collated_dataset_path, format="parquet")
+    collated_df = collated_dataset.to_table().to_pandas()
+
+    assert collated_dataset.files == [
+        str(Path(collated_dataset_path) / "records-0.parquet")
+    ]
+    assert len(collated_df) == 10
+    assert set(collated_df["source"].unique()) == {"MIT Alma", "DSpace@MIT"}
```
ghukill, line 53: I like it!  Nice assertions here.

ghukill (approved):
Thanks for all the work and updates on this pretty foundational part of the app!  Looks good to me.

ehanson8 (approved):
Fantastic work on this, approved!

PR #31, opened 2024-10-18
TIMX 372- support runs with zero diffs

ghukill:

A bug surfaced where attempting to calculate metrics from the diffs dataset would throw an error if zero records had any diffs.  This was the result of dynamically preparing SQL queries for DuckDB that looked for any field with a value of `1`. With no diffs, there were no TIMDEX field columns, and the 'OR' clause was malformed.

In the process of addressing this bug, it was determined that this somewhat brittle SQL query generation could be avoided by carrying over the `has_diff` column from the diffs dataset and use that to identify rows in the sparse matrix that had some kind of diff.

Some updates were needed to the webapp as well to gracefully handle runs with zero diffs.  One part was updating the queries to use this new `has_diff` column in the sparse matrix to identify records with diffs (vs looking at TIMDEX field name columns).  Additionally, to indicate to the user quite obviously when a run has no diffs.

Banner added to top of page indicating no diffs:

Screenshot 2024-10-18 at 1 57 22 PM

And the "Record Samples" section updated to more clearly indicate either situation:

runs with diffs:
Screenshot 2024-10-18 at 1 57 01 PM

runs without diffs:
Screenshot 2024-10-18 at 1 57 17 PM

TODO....

YES | NO

YES | NO

- Include links to Jira Software and/or Jira Service Management tickets here.

### Developer
- [ ] All related Jira tickets are linked in commit message(s)
- [ ] Stakeholder approval has been confirmed (or is not needed)

### Code Reviewer(s)

### DIFF
--- a/abdiff/core/calc_ab_metrics.py
+++ b/abdiff/core/calc_ab_metrics.py
@@ -9,14 +9,13 @@ from pathlib import Path
 import duckdb
 import pandas as pd
 import pyarrow as pa
-import pyarrow.dataset as ds
 from duckdb.duckdb import DuckDBPyConnection
 
-from abdiff.core.utils import update_or_create_run_json, write_to_dataset
+from abdiff.core.utils import load_dataset, update_or_create_run_json, write_to_dataset
 
 logger = logging.getLogger(__name__)
 
-NON_TIMDEX_FIELD_COLUMNS = ["timdex_record_id", "source"]
+NON_TIMDEX_FIELD_COLUMNS = ["timdex_record_id", "source", "has_diff"]
 
 
 def calc_ab_metrics(
@@ -52,20 +51,22 @@ def create_record_diff_matrix_dataset(
     This writes a single parquet file with rows for each record, and columns for each
     TIMDEX field, and a value of integer 1 if that field has a diff and 0 if not.  This
     provides a handy way to calculate aggregate metrics for a given field or source in
-    later steps.
+    later steps.  The column "has_diff" is also carried over from the diffs dataset to
+    provide a single column to check if ANY of the field columns indicate a diff for a
+    record row.
 
     This code momentarily creates a single dataframe in memory for all rows.  This is safe
     given the nature of the dataframe: there may be 10m rows, and potentially 20-30
     columns, but all values are either integer 1 or 0 representing a boolean, resulting in
     a very small dataset in memory despite a potentially high row count.
     """
-    diffs_ds = ds.dataset(diffs_dataset)
+    diffs_ds = load_dataset(diffs_dataset)
 
     batch_metrics_dfs = []
     for i, batch in enumerate(
         diffs_ds.to_batches(
             batch_size=batch_size,
-            columns=["timdex_record_id", "source", "ab_diff"],
+            columns=["timdex_record_id", "source", "ab_diff", "has_diff"],
         )
     ):
         start_time = time.time()
@@ -81,6 +82,7 @@ def create_record_diff_matrix_dataset(
             record_metrics = {
                 "timdex_record_id": row["timdex_record_id"],
                 "source": row["source"],
+                "has_diff": 1 if row["has_diff"] == "true" else 0,
             }
             diff_data = row["ab_diff"]
             record_metrics.update(generate_field_diff_bools_for_record(diff_data))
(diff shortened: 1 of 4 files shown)


abdiff/core/calc_ab_metrics.py
```diff
@@ -238,12 +239,12 @@ def _get_source_counts(
     sources: list[str],
     analysis: dict,
 ) -> dict:
-    any_field_modified_condition = " OR ".join(f"{field} = 1" for field in fields)
```
ghukill, line 241: This is indicative of the key change from the PR.  We no longer need to dynamically construct a SQL query to look for a True `1` value for _any_ TIMDEX field, field names we do not know in advance.

Instead, we can just apply the simple, static clause `where has_diff` in its place anywhere we want to identify records with diffs.

ehanson8 (approved):
Good work, this is a much cleaner approach in my mind!

abdiff/core/calc_ab_metrics.py
```diff
@@ -215,17 +217,16 @@ def _prepare_duckdb_context(
...
         SELECT COUNT(*) FROM record_diff_matrix
         """
     ).fetchone()[0]
 
-    any_field_modified_condition = " OR ".join(f"{field} = 1" for field in fields)
     total_records_with_diff = conn.execute(
-        f"""
-        select count(*) from record_diff_matrix where {any_field_modified_condition}
+        """
+        select count(*) from record_diff_matrix where has_diff;
```
jonavellecuerdo, line 229: Hmm, should this query read

```sql
select count(*) from record_diff_matrix where has_diff=1
```
based on the [derivation for `has_diff`](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/31/files#diff-9080a340ec2806075e422a9e58284e7dee2a0c7ccf5f4ff2d8178c03ec5e1c97R85)?

abdiff/core/calc_ab_metrics.py
```diff
@@ -238,12 +239,12 @@ def _get_source_counts(
...
     analysis: dict,
 ) -> dict:
-    any_field_modified_condition = " OR ".join(f"{field} = 1" for field in fields)
     for source in sources:
         total_count = conn.execute(
             f"""
             select count(*) from record_diff_matrix
-            where source = '{source}' and ({any_field_modified_condition})
+            where source = '{source}'
+            and has_diff;
```
jonavellecuerdo, line 247: Same question here re: whether query should instead read:

```sql
has_diff = 1
```

abdiff/core/calc_ab_metrics.py
ghukill: Your right that `had_diff` is getting either the integer `1` or `0`, but you can treat this as a boolean when querying DuckDB (and most DBs).

e.g.

```

-- 1 is considered TRUE
D select 1 == true;
┌────────────────────────────┐
│ (1 = CAST('t' AS BOOLEAN)) │
│          boolean           │
├────────────────────────────┤
│ true                       │
└────────────────────────────┘
D select 0 == true;
┌────────────────────────────┐
│ (0 = CAST('t' AS BOOLEAN)) │
│          boolean           │
├────────────────────────────┤
│ false                      │
└────────────────────────────┘

-- 0 is considered FALSE
D select 0 == true;
┌────────────────────────────┐
│ (0 = CAST('t' AS BOOLEAN)) │
│          boolean           │
├────────────────────────────┤
│ false                      │
└────────────────────────────┘
D select 0 == false;
┌────────────────────────────┐
│ (0 = CAST('f' AS BOOLEAN)) │
│          boolean           │
├────────────────────────────┤
│ true                       │
└────────────────────────────┘
```

So, safe to just query like `where has_diff` which implies `where has_diff = 1`.

abdiff/core/calc_ab_metrics.py
ghukill: Same answer!

abdiff/core/calc_ab_metrics.py
jonavellecuerdo: Hmm, is the "casting to boolean" default behavior? 🤔

ghukill: @jonavellecuerdo - to my knowledge, yes it is.  I don't lean into this convention all that often, preferring the explicitness of true boolean values where possible (though.... they likely just store a `1` or `0` under the hood in the given system).  But in this sparse matrix, where conversions between python / pyarrow / parquet / duckdb could be subtly change types, a simple integer felt ideal.

The same is even true in python, further demonstrating it's kind of a "default" behavior in many systems:

```
In [1]: assert 1 == True

In [2]: assert 1 == False
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 assert 1 == False

AssertionError: 

In [3]: assert 0 == True
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 assert 0 == True

AssertionError: 

In [4]: assert 0 == False
```

PR #35, opened 2024-10-22
Timx 369 core status checks

jonavellecuerdo:
Implement integrity checks by validating the outputs generated by core functions called in the CLI command `run-diff`. The core functions in `run-diff` this PR focuses on are: 
* `run_ab_transforms`
* `collate_ab_transforms`

These two functions have been updated to raise exceptions with the goal of stopping the flow of execution when invalid outputs are generated (as opposed to proceeding to call downstream functions that then fail as a result of invalid inputs). This will help us to better identify what causes an A/B diff run to fail.

The updates are summarized below: 
* `run_ab_transforms`
   * [Raise an error if any containers exited **with an error**.](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/35/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR117-R119) The error message will list the IDs of the failing containers, which have been added to the log file (it is now the header for each "section" of logs, where each "section"= a container run).
      * This required making updates to the function `wait_for_containers` to [check the exit code of an exited container](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/35/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR198-R206), which should be 0 if it ran successfully and 1 if an error was encountered. The function `wait_for_containers` will now return a list of IDs for failed containers; if all containers ran successfully, the function returns `None`.
      
      **Note:** The reason for doing it this way is to identify _all_ failing containers and not raise an error as soon as one container fails. It might be worth noting that if containers **are not** run in detached mode, the `docker` module would raise a `docker.errors.ContainerError` (you can read more about this error in the [docs](https://docker-py.readthedocs.io/en/stable/containers.html#container-objects)).
   * [Raise an error if the **length** of any of the transformed file _lists_ from `get_transformed_files` does not match the **length** of `input_files`.](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/35/files#diff-7a49d9dbf619fca67eb5ed5ed16b6c62440a4cc6898754cd835d2b819be74ddcR259-R267)
* `collate_ab_transforms`
   * [Raise an error if the number of rows in the collated dataset is zero (i.e., empty). ](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/35/files#diff-1cf5419c409425ac3b8275acb901b7c68cc1131a5b24d5846f477eb967681e55R224-R228)
   * [Raise an error if at least one of the `record_*` columns is all NA/null (i.e., empty).](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/35/files#diff-1cf5419c409425ac3b8275acb901b7c68cc1131a5b24d5846f477eb967681e55R230-R236)

Additional update(s): 
* `build_ab_images`
   * [Raise an error if attempting to reset cloned repository to an invalid commit SHA. 
](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/35/files#diff-a05d142e2388796ded9aed68f6d4481d3e3021603b6681fdc32d631bf81311b9R99-R105).

1. Review unit tests and confirm all tests are passing.
2. Run the following command below to see the error that gets raised when an invalid commit SHA is passed to `build_ab_images`: 
   ```shell
   pipenv run abdiff --verbose init-job -d output/test-job -m "test job" -a abc123 -b def456
   ```
   
   The screenshot below depicts the output you should see in your terminal.
   image

3. Clear out your `.env` file and run the following command to see what happens when `run_ab_transforms` is executed without AWS credentials (i.e., container exits with an error).
   * Run `init-job`: 
      ```shell
      pipenv run abdiff --verbose init-job \
         -d output/mvp \
         -m "test of full CLI workflow" \
         -a 395e612 \
         -b cf1024c
      ```
   * Run `run-diff`: 
      ```shell
      pipenv run abdiff --verbose run-diff \
         -d output/mvp \
         -i "s3://timdex-extract-prod-300442551476/alma/alma-2023-02-19-daily-extracted-records-to-index.xml,s3://timdex-extract-prod-300442551476/libguides/libguides-2024-04-03-full-extracted-records-to-index.xml,s3://timdex-extract-prod-300442551476/dspace/dspace-2024-10-11-daily-extracted-records-to-index.xml"
      ```

   The screenshot below depicts the output you should see in your terminal. 
   image

   **Note:** Trying to force raise the other errors implemented in this update would be a bit tricky, and the unit tests should suffice for this review. 🤓 

YES - `Pipfile.lock` previously did not include the updated set of dependencies in `Pipfile`. 

NO

- https://mitlibraries.atlassian.net/browse/TIMX-369

### Developer

### Code Reviewer(s)

### DIFF
--- a/abdiff/core/build_ab_images.py
+++ b/abdiff/core/build_ab_images.py
@@ -7,6 +7,7 @@ import docker.models.images
 from pygit2 import clone_repository
 from pygit2.enums import ResetMode
 
+from abdiff.core.exceptions import InvalidRepositoryCommitSHAError
 from abdiff.core.utils import update_or_create_job_json
 
 logger = logging.getLogger(__name__)
@@ -92,10 +93,17 @@ def clone_repo_and_reset_to_commit(clone_directory: str, commit_sha: str) -> Non
         commit_sha: The SHA of a repo commit.
     """
     logger.debug(f"Cloning repo to: {clone_directory}")
+    transmogrifier_url = "https://github.com/MITLibraries/transmogrifier.git"
     repository = clone_repository(
-        "https://github.com/MITLibraries/transmogrifier.git",
+        transmogrifier_url,
         clone_directory,
     )
     logger.debug(f"Cloned repo to: {clone_directory}")
-    repository.reset(commit_sha, ResetMode.HARD)
-    logger.debug(f"Cloned repo reset to commit: {commit_sha}")
+
+    try:
+        repository.reset(commit_sha, ResetMode.HARD)
+        logger.debug(f"Cloned repo reset to commit: {commit_sha}")
+    except KeyError as exception:
+        raise InvalidRepositoryCommitSHAError(
+            transmogrifier_url, commit_sha
+        ) from exception
--- a/abdiff/core/collate_ab_transforms.py
+++ b/abdiff/core/collate_ab_transforms.py
@@ -9,7 +9,6 @@ from pathlib import Path
 import duckdb
 import ijson
 import pyarrow as pa
-import pyarrow.dataset as ds
 
 from abdiff.core.exceptions import OutputValidationError
 from abdiff.core.utils import write_to_dataset
@@ -168,7 +167,7 @@ def get_joined_batches_iter(dataset_directory: str) -> Generator[pa.RecordBatch]
         dataset_directory: The root directory of the Parquet dataset of TIMDEX records
             (i.e., the tempfile.TemporaryDirectory).
     """
-    with duckdb.connect() as con:
+    with duckdb.connect(":memory:") as con:
         transformed_file_names = con.execute(
             """
             SELECT DISTINCT transformed_file_name FROM
@@ -216,24 +215,51 @@ def validate_output(dataset_path: str) -> None:
     """Validate the output of collate_ab_transforms.
 
     This function checks whether the collated dataset is empty
-    and whether any or both 'record_*' columns are empty.
+    and whether any or both 'record_a' or 'record_b' columns are
+    totally empty.
     """
-    collated_dataset = ds.dataset(dataset_path)
-    collated_table = collated_dataset.to_table(columns=["record_a", "record_b"])
+    collated_parquet_glob = f"{dataset_path}/**/*.parquet"
 
-    # check if the table is empty
-    if collated_table.num_rows == 0:
-        raise OutputValidationError(  # noqa: TRY003
-            "The collated dataset does not contain any records."
-        )
+    with duckdb.connect(":memory:") as con:
+        # check if the table is empty
+        record_count = con.execute(
+            "SELECT COUNT(*) FROM read_parquet($collated_parquet_glob)",
+            {"collated_parquet_glob": collated_parquet_glob},
+        ).fetchone()[
+            0
+        ]  # type: ignore[index]
 
-    # check if any of the 'record_*' columns are empty
-    record_columns = collated_table.to_pandas()
-    if record_columns["record_a"].isna().all() or record_columns["record_b"].isna().all():
-        raise OutputValidationError(  # noqa: TRY003
-            "At least one or both record column(s) ['record_a', 'record_b'] "
-            "in the collated dataset are empty."
-        )
+        if record_count == 0:
+            raise OutputValidationError(  # noqa: TRY003
+                "The collated dataset does not contain any records."
+            )
+
+        # check if any of the 'record_*' columns are empty
+        record_a_null_count = con.execute(
+            """
+            SELECT COUNT(*) FROM read_parquet($collated_parquet_glob)
+            WHERE record_a ISNULL
+            """,
+            {"collated_parquet_glob": collated_parquet_glob},
+        ).fetchone()[
+            0
+        ]  # type: ignore[index]
+
+        record_b_null_count = con.execute(
+            """
+            SELECT COUNT(*) FROM read_parquet($collated_parquet_glob)
+            WHERE record_b ISNULL
+            """,
+            {"collated_parquet_glob": collated_parquet_glob},
+        ).fetchone()[
+            0
+        ]  # type: ignore[index]
+
+        if record_count in {record_a_null_count, record_b_null_count}:
+            raise OutputValidationError(  # noqa: TRY003
+                "At least one or both record column(s) ['record_a', 'record_b'] "
+                "in the collated dataset are empty."
+            )
 
 
 def parse_parquet_details_from_transformed_file(transformed_file: str) -> tuple[str, ...]:
--- a/abdiff/core/collate_ab_transforms.py
+++ b/abdiff/core/collate_ab_transforms.py
@@ -9,7 +9,9 @@ from pathlib import Path
 import duckdb
 import ijson
 import pyarrow as pa
+import pyarrow.dataset as ds
 
+from abdiff.core.exceptions import OutputValidationError
 from abdiff.core.utils import write_to_dataset
 
 logger = logging.getLogger(__name__)
@@ -67,6 +69,9 @@ def collate_ab_transforms(
         schema=JOINED_DATASET_SCHEMA,
     )
     logger.info(f"Wrote {len(joined_written_files)} parquet file(s) to collated dataset")
+
+    validate_output(collated_dataset_path)
+
     return collated_dataset_path
 
 
@@ -187,8 +192,8 @@ def get_joined_batches_iter(dataset_directory: str) -> Generator[pa.RecordBatch]
                     a AS (SELECT * FROM transformed_file WHERE version='a'),
                     b AS (SELECT * FROM transformed_file WHERE version='b')
                 SELECT
-                    a.timdex_record_id,
-                    a.source,
+                    COALESCE(a.timdex_record_id, b.timdex_record_id) timdex_record_id,
+                    COALESCE(a.source, b.source) source,
                     a.record as record_a,
                     b.record as record_b
                 FROM a
@@ -205,6 +210,30 @@ def get_joined_batches_iter(dataset_directory: str) -> Generator[pa.RecordBatch]
                     yield results.read_next_batch()
                 except StopIteration:
                     break
+
+
+def validate_output(dataset_path: str) -> None:
+    """Validate the output of collate_ab_transforms.
+
+    This function checks whether the collated dataset is empty
+    and whether any or both 'record_*' columns are empty.
+    """
+    collated_dataset = ds.dataset(dataset_path)
+    collated_table = collated_dataset.to_table(columns=["record_a", "record_b"])
+
+    # check if the table is empty
+    if collated_table.num_rows == 0:
+        raise OutputValidationError(  # noqa: TRY003
+            "The collated dataset does not contain any records."
+        )
+
+    # check if any of the 'record_*' columns are empty
+    record_columns = collated_table.to_pandas()
+    if record_columns["record_a"].isna().all() or record_columns["record_b"].isna().all():
+        raise OutputValidationError(  # noqa: TRY003
+            "At least one or both record column(s) ['record_a', 'record_b'] "
+            "in the collated dataset are empty."
+        )
 
 
 def parse_parquet_details_from_transformed_file(transformed_file: str) -> tuple[str, ...]:
--- a/abdiff/core/collate_ab_transforms.py
+++ b/abdiff/core/collate_ab_transforms.py
@@ -218,17 +218,18 @@ def validate_output(dataset_path: str) -> None:
     and whether any or both 'record_a' or 'record_b' columns are
     totally empty.
     """
-    collated_parquet_glob = f"{dataset_path}/**/*.parquet"
-
     with duckdb.connect(":memory:") as con:
-        # check if the table is empty
-        record_count = con.execute(
-            "SELECT COUNT(*) FROM read_parquet($collated_parquet_glob)",
-            {"collated_parquet_glob": collated_parquet_glob},
-        ).fetchone()[
-            0
-        ]  # type: ignore[index]
+        # create view of collated table
+        con.execute(
+            f"""
+            CREATE VIEW collated AS (
+                SELECT * FROM read_parquet('{f"{dataset_path}/**/*.parquet"}')
+            )
+            """
+        )
 
+        # check if the table is empty
+        record_count = con.execute("SELECT COUNT(*) FROM collated").fetchone()[0]  # type: ignore[index]
         if record_count == 0:
             raise OutputValidationError(  # noqa: TRY003
                 "The collated dataset does not contain any records."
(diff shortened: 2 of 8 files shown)


abdiff/core/build_ab_images.py
```diff
@@ -92,10 +93,17 @@ def clone_repo_and_reset_to_commit(clone_directory: str, commit_sha: str) -> Non
...
-    repository.reset(commit_sha, ResetMode.HARD)
-    logger.debug(f"Cloned repo reset to commit: {commit_sha}")
+
+    try:
+        repository.reset(commit_sha, ResetMode.HARD)
+        logger.debug(f"Cloned repo reset to commit: {commit_sha}")
+    except KeyError as exception:
+        raise InvalidRepositoryCommitSHAError(
+            transmogrifier_url, commit_sha
+        ) from exception
```
ghukill, line 109: +1 - though not technically part of CLI command `run-diff`, this feels like a nice check to have, and will bubble up naturally in CLI command `init-job`.

abdiff/core/run_ab_transforms.py
```diff
@@ -101,22 +108,31 @@ def run_ab_transforms(
...
 
-    wait_for_containers(containers, timeout)
+    failed_containers = wait_for_containers(containers, timeout)
     logger.info(f"All {len(containers)} containers have exited.")
 
     log_file = aggregate_logs(run_directory, containers)
     logger.info(f"Log file created: {log_file}")
 
-    transformed_files = get_transformed_files(run_directory)
+    # errors from failed containers are accessible via aggregated logs
```
ghukill, line 117: I think this is a really nice decision.  By aggregating the logs from the containers _before_ raising an exception that something went wrong, we have an enduring artifact about what happened.  This could be invaluable if we are running 10, 20, 50 input files through.

abdiff/core/run_ab_transforms.py
```diff
@@ -155,13 +171,21 @@ def run_docker_container(
     return container
 
 
-def wait_for_containers(containers: list[Container], timeout: int = 180) -> None:
-    """Given a list of running containers, wait until all containers exit."""
-    exited_containers: list[Container] = []
+def wait_for_containers(
```
ghukill, line 174: I'm going to comment kind of high level on this function.

I have some local spike code for [TIMX-356](https://mitlibraries.atlassian.net/browse/TIMX-356) which seeks to limit how many containers can run in parallel.  This will necessarily revise a bit of the signatures of these functions, where we may choose to return a tuple like `([completed containers], [container exceptions])` from this function.

That said, I think this approach makes sense in this context.  And with this change in place, I will work to interleave this approach with any changes from TIMX-356. 

Interestingly (and encouragingly), I think the signature of this function is _more_ like what I had envisioned in TIMX-356, so feeling good about approving this as-is.

abdiff/core/run_ab_transforms.py
```diff
@@ -171,21 +195,34 @@ def wait_for_containers(containers: list[Container], timeout: int = 180) -> None
...
+    return None
+
 
 def aggregate_logs(run_directory: str, containers: list[Container]) -> str:
     """Retrieve logs for containers in a list, aggregating to a single log file."""
     log_file = str(Path(run_directory) / "transformed/logs.txt")
     with open(log_file, "w") as file:
         for container in containers:
-            header = (
+            file.write(f"container: {container.id}\n")
```
ghukill, line 219: Nice addition to include the container as a section header.

abdiff/core/run_ab_transforms.py
```diff
@@ -211,6 +248,25 @@ def get_transformed_files(run_directory: str) -> tuple[list[str], ...]:
...
+) -> None:
+    """Validate the output of run_ab_transforms.
+
+    This function checks that the number of files in the A/B transformed file
+    directories matches expected file count (i.e., the number of input files).
+    """
+    if any(
+        len(transformed_files) != expected_file_count
+        for transformed_files in ab_transformed_file_lists
+    ):
```
ghukill, line 262: I must admit I had to read this a couple times to understand how and why it worked, but I think I get it!  

My confusion came from `expected_file_count` which I understand now to be basically the lenght of the list of input files?  So if 5 input files were passed, we'd expect _both_ lists in `ab_transformed_file_lists` to be length 5.  

I had originally thought, "if we have 5 input files, wouldn't the `expected_file_count` be 10?  for both "a" and "b" versions?"  which led me to incorrectly wonder why the individual lists would equal that length.

What about a minor variable renaming?  Something like:

```python
if any(
    len(version_transformed_files) != input_files_count
    for version_transformed_files in ab_transformed_file_lists
)
```

or, just thinking about readability and explicitness, maybe another option:

```python
a_transformed_files, b_transformed_files = ab_transformed_file_lists
if (
    len(a_transformed_files) != input_files_count
    or len(b_transformed_files) != input_files_count
):
  # one of them is off....
```

Feel free to disregard, not a requested change, just an option.

abdiff/core/collate_ab_transforms.py
```diff
@@ -207,6 +212,30 @@ def get_joined_batches_iter(dataset_directory: str) -> Generator[pa.RecordBatch]
                     break
 
 
+def validate_output(dataset_path: str) -> None:
+    """Validate the output of collate_ab_transforms.
+
+    This function checks whether the collated dataset is empty
+    and whether any or both 'record_*' columns are empty.
```
ghukill, line 219: Could rewrite as,

> or 'record_a' or 'record_b' columns are totally empty.

abdiff/core/collate_ab_transforms.py
```diff
@@ -207,6 +212,30 @@ def get_joined_batches_iter(dataset_directory: str) -> Generator[pa.RecordBatch]
...
+    """
+    collated_dataset = ds.dataset(dataset_path)
+    collated_table = collated_dataset.to_table(columns=["record_a", "record_b"])
```
ghukill, line 222: I hate to say it, but I think this might need some rework.  Calling `collated_dataset.to_table(...)` will pull the entire dataset into memory.

I think checking if the dataset is entirely empty could be handled by dataset methods like `.take(10)` or `.head(10)`, as you'd expect those to return zero results if so, but it's only getting the first few records which is memory safe.

As for checking if `record_a` or `record_b` is totally empty... it's possible that a DuckDB query could be an efficiente way to do this?

In fact, DuckDB query(s) could be used for _all_ of this, with the expectation that DuckDB is memory efficient and would not load the entire dataset into memory.

ghukill (changes requested):
Overall, I think this is very nicely done, and drastically help bubble up errors when running the CLI command `run-diff`.  The overall pattern makes good sense to me.

Left a couple of stylistic comments, which are totally optional.

But the "Request changes" decision comes from the `collate_ab_transforms.validate_output()` function, which I think runs the risk of pulling the entire dataset into memory.  Happy to discuss more if you'd like, but I do think that needs some reworking to not overwhelm memory for large runs.

abdiff/core/run_ab_transforms.py
ehanson8: Agreed!

tests/test_build_ab_images.py
```diff
@@ -29,6 +32,24 @@ def test_build_ab_images_success(mocked_clone, job_directory, mocked_docker_clie
...
+
+    with pytest.raises(InvalidRepositoryCommitSHAError):
+        build_ab_images(
+            job_directory,
+            "invalid",
+            "def456",
+            mocked_docker_client,
+        )
+
+
```
ehanson8, line 52: Great test!

abdiff/core/run_ab_transforms.py
ehanson8: I agree with `input_files_count` as a variable name, `expected` has more ambiguity

tests/test_run_ab_transforms.py
```diff
@@ -50,6 +56,32 @@ def test_run_ab_transforms_success(
         )
 
 
+def test_run_ab_transforms_raise_error(
```
ehanson8, line 59: I'd be clearer about what is triggering the error in the test name

abdiff/core/collate_ab_transforms.py
ehanson8: ☝️

abdiff/core/collate_ab_transforms.py
ehanson8: This wouldn't have been obvious to me either, so thank you very much for pointing it out!

ehanson8 (commented):
Great work, some minor suggestions and I do agree with @ghukill's request to rework `validate_output`

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Hooray! It'd be great if we could parse only the information we need (i.e., the error from the logs) or create a separate log file for errors, but I think this is sufficient for our first pass!

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Thank you for being flexible! This is helpful to know and I'm looking forward to reviewing the updates for TIMX-356. 🤓

abdiff/core/run_ab_transforms.py
jonavellecuerdo: Ah! Thank you for this suggestion. I updated the parameter name and the function docstring.

tests/test_run_ab_transforms.py
jonavellecuerdo: Renamed test to `test_run_ab_transforms_raise_error_if_containers_failed`.

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: @ghukill Just wanted to clarify (and make sure I understand as well): 
The code highlighted above would load only the **selected columns** specified in `.to_table(columns=[...])`, right? From the [PyArrow docs](https://arrow.apache.org/docs/python/dataset.html#dataset-discovery) it reads:

> Creating a [Dataset](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Dataset.html#pyarrow.dataset.Dataset) object does not begin reading the data itself. If needed, it only crawls the directory to find all the files:

However, what you're saying is that even if we are only loading in a subset of columns `record_a` and `record_b`, that can _still_ consume a lot of memory if we are diff-ing a lot of input files. 

Let me know what ya'll think of [the update to use DuckDB queries](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/35/files#diff-1cf5419c409425ac3b8275acb901b7c68cc1131a5b24d5846f477eb967681e55R221-R262). I think I personally prefer the concise syntax with the previous method, _but_ I definitely want to be mindful of memory use!

abdiff/core/collate_ab_transforms.py
```diff
@@ -216,24 +215,51 @@ def validate_output(dataset_path: str) -> None:
...
+    totally empty.
     """
-    collated_dataset = ds.dataset(dataset_path)
-    collated_table = collated_dataset.to_table(columns=["record_a", "record_b"])
+    collated_parquet_glob = f"{dataset_path}/**/*.parquet"
+
+    with duckdb.connect(":memory:") as con:
+        # check if the table is empty
+        record_count = con.execute(
+            "SELECT COUNT(*) FROM read_parquet($collated_parquet_glob)",
```
ghukill, line 226: Commenting here, but applies to pattern of the queries.

Another option when we have a DuckDB connection, assuming we reuse it for multiple queries, is that we could create a view that we reuse in other queries, e.g.

```python
con.execute(
    """
    CREATE VIEW collated as 
    SELECT * FROM read_parquet($collated_parquet_glob,hive_partitioning=true)
    """,
    {"collated_parquet_glob": collated_parquet_glob},
)
# no return value needed, as it creates a database object used in future queries
```

With that view created, which lasts for the duration of the connection, you could write queries like the following, no longer needing to repeat the `read_parquet(...)`:

```sql
SELECT COUNT(*) FROM collated;
```

100% optional.  Approving regardless of change, and no need for review if changed!

ghukill (approved):
Looks good to me!

Thanks for the small changes, and the update to `validate_output` makes sense to me.  Where possible, I think leaning into DuckDB's memory management can be quite helpful, as in this example.  While still possible to write queries that would be memory prohibitive, even then, DuckDB may "spill to disk" and be slower, but would not fail.

abdiff/core/collate_ab_transforms.py
ghukill: You're right that loading a dataset does not bring anything into memory until other operations are performed, 100%.  So we're safe until the `to_table()` call.

> However, what you're saying is that even if we are only loading in a subset of columns record_a and record_b, that can still consume a lot of memory if we are diff-ing a lot of input files.

That's also correct!  But, keep in mind that `record_a` and `record_b` are binary blobs of the _full_ records, so they are hold relatively large values per row.  

Assume the collate dataset held the A and B versions of records from `alma`, maybe around 3 million records.  The transformed records from A or B (just one) is about 14gb in JSON files, probably somewhere around 10gb in parquet (compressed).  Regardless of compression, bring in _both_ A and B versions (`record_a` and `record_b` columns), which be potentially around 20gb in memory.  Not only would this be quite slow, it would often overwhelm memory (I believe all of our machines are 16gb).  

To summarize: even though we subset the columns, and it's lazy loaded until `to_table()`, the size of those columns makes it an very memory-expensive operation.

Interestingly, when calcing metrics, we _do_ call something similar to `to_table()` on the entire dataset, but we are filtering down to columns with very, very small values.  So even if 10-20 million rows, it may only be 50-100mb in memory, tops.  The nature of the columns we select are critical to deciding if we can safely bring into memory of not.

ehanson8 (approved):
Looks great!

abdiff/core/collate_ab_transforms.py
jonavellecuerdo: @ehanson8 @ghukill As discussed:
* Updated validation function to create a view of the collated table that gets used by the subsequent queries.

Also worth noting: 
* DuckDB does not support using prepared statements when creating views (seems to be a [known issue](https://github.com/duckdb/duckdb/issues/13069))
* Can't use parameters in place of column names in query

abdiff/core/collate_ab_transforms.py
```diff
@@ -218,39 +218,32 @@ def validate_output(dataset_path: str) -> None:
...
-            "SELECT COUNT(*) FROM read_parquet($collated_parquet_glob)",
-            {"collated_parquet_glob": collated_parquet_glob},
-        ).fetchone()[
-            0
-        ]  # type: ignore[index]
+        # create view of collated table
+        con.execute(
+            f"""
+            CREATE VIEW collated AS (
+                SELECT * FROM read_parquet('{f"{dataset_path}/**/*.parquet"}')
```
ghukill, line 226: I think this will work given that we don't need any partitions from the dataset -- in fact, don't think there _are_ partitions -- but worth noting that if we _did_, we'd need to add `hive_partitioning=true`. 

Again, not needed here, just noting!

ghukill (approved):
Approved!  And nice catch about the inability to use named parameters in the create view syntax, and instead using the f-string.

ehanson8 (approved):

PR #45, opened 2024-11-01
TIMX 379 - CSV for input files support and helpers

ghukill:

This PR partially addresses [TIMX-379](https://mitlibraries.atlassian.net/browse/TIMX-379) which is concerned with performing large / complex runs.

**The goal**: perform a run of ABDiff where the input files are for ALL sources in TIMDEX, going back to their most recent "full" run, and then all subsequent "daily" runs.

This full list of input files is around 1.6k (very many!) and exposes two challenges:
1. creating this list of input files
2. passing these to the CLI command `run-diff` as comma seperated values for `-i / --input-files`

**This PR addresses this in two ways**:
1. create a new CLI command `timdex-sources-csv` that will generate a CSV of input files based on some criteria
2. updates CLI command `run-diff` to support a CSV for the `-i / --input-files` argument for input files

I would highlight a couple of characteristics of these changes:
- they are mostly purely additive, meaning no original logic or behavior is changed
- they are not well tested

I anticipate that the "not well tested" part makes some nervous, but as purely additive changes for just initiating runs, I would propose that's okay.  These changes support larger runs for testing at scale, which in turn focus and direct work on bugs and features that do warrant more rigourous testing. 

1- Create a job
```shell
pipenv run abdiff --verbose init-job -d output/jobs/csv -a 008e20c -b 395e612
```

2- Set AWS credentials for production in `.env` file

3- Set additional env var required for CLI command `timdex-sources-csv`:
```text
TIMDEX_BUCKET=timdex-extract-prod-300442551476
```

4- Run new CLI command to generate a CSV file of input files for `libguides` and `gismit` sources
```shell
pipenv run abdiff --verbose timdex-sources-csv \
-o output/jobs/csv/test.csv \
-s libguides,gismit
```
- observe that new CSV file created at `output/jobs/csv/test.csv` contains ~182 S3 URIs
- files are ordered by last "full" run then all following "daily" runs

5- Run diff and pass CSV
```shell
pipenv run abdiff --verbose run-diff \
-d output/jobs/csv \
-i output/jobs/csv/test.csv
```
- observe that run uses CSVs
- though 182 input files (so 364 containers), runs pretty quick given small input file sizes, about 2-3 minutes total time

YES

NO

- https://mitlibraries.atlassian.net/browse/TIMX-379

### Developer

### Code Reviewer(s)

[TIMX-379]: https://mitlibraries.atlassian.net/browse/TIMX-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ WEBAPP_HOST=# host for flask webapp
 WEBAPP_PORT=# port for flask webapp
 TRANSMOGRIFIER_MAX_WORKERS=# max number of Transmogrifier containers to run in parallel; default is 6
 TRANSMOGRIFIER_TIMEOUT=# timeout for a single Transmogrifier container; default is 5 hours
+TIMDEX_BUCKET=# when using CLI command 'timdex-sources-csv', this is required to know what TIMDEX bucket to use
 ```
 
 ## CLI commands
@@ -143,7 +144,10 @@ Usage: -c run-diff [OPTIONS]
 
 Options:
   -d, --job-directory TEXT  Job directory to create.  [required]
-  -i, --input-files TEXT    Input files to transform.  [required]
+  -i, --input-files TEXT    Input files to transform.  This may be a comma
+                            separated list of input files, or a local CSV file
+                            that provides a list of files.  [required]
+  -m, --message TEXT        Message to describe Run.
   -h, --help                Show this message and exit.
 ```
 
@@ -158,3 +162,23 @@ Options:
   -h, --help                Show this message and exit.
 ```
 
+### `timdex-sources-csv`
+```text
+Usage: -c timdex-sources-csv [OPTIONS]
+
+  Generate a CSV of ordered extract files for all, or a subset, of TIMDEX
+  sources.
+
+  This CSV may be passed to CLI command 'run-diff' for the '-i / --input-
+  files' argument, serving as the list of input files for the run.
+
+  This command requires that env var 'TIMDEX_BUCKET' is set to establish what
+  S3 bucket to use for scanning.  The appropriate AWS credentials are also
+  needed to be set.
+
+Options:
+  -o, --output-file TEXT  Output filepath for CSV.  [required]
+  -s, --sources TEXT      Optional comma separated list of sources to include.
+                          Default is all.
+  -h, --help              Show this message and exit.
+```
--- a/abdiff/helpers/timdex_sources.py
+++ b/abdiff/helpers/timdex_sources.py
@@ -29,7 +29,7 @@ def get_extracted_files_for_source(
             for obj in page["Contents"]:
                 if not obj["Key"].endswith("/"):  # skip folders
                     s3_uri = f"s3://{bucket}/{obj['Key']}"
-                    files.append(s3_uri)  # noqa: PERF401
+                    files.append(s3_uri)
 
     # filter where "extracted" in filename
     return [file for file in files if "extracted" in file]
--- a/abdiff/helpers/timdex_sources.py
+++ b/abdiff/helpers/timdex_sources.py
@@ -1,0 +1,93 @@
+"""abdiff.helpers.timdex_sources"""
+
+import datetime
+import logging
+import re
+
+import boto3  # type: ignore[import-untyped]
+
+from abdiff.config import Config
+
+logger = logging.getLogger(__name__)
+
+CONFIG = Config()
+
+
+def get_extracted_files_for_source(
+    source: str,
+    bucket: str = CONFIG.TIMDEX_BUCKET,
+) -> list[str]:
+    """List S3 URIs for extract files in TIMDEX S3 bucket for a given source."""
+    s3_client = boto3.client("s3")
+    files = []
+
+    paginator = s3_client.get_paginator("list_objects_v2")
+    page_iterator = paginator.paginate(Bucket=bucket, Prefix=source)
+
+    for page in page_iterator:
+        if "Contents" in page:
+            for obj in page["Contents"]:
+                if not obj["Key"].endswith("/"):  # skip folders
+                    s3_uri = f"s3://{bucket}/{obj['Key']}"
+                    files.append(s3_uri)  # noqa: PERF401
+
+    # filter where "extracted" in filename
+    return [file for file in files if "extracted" in file]
+
+
+def get_ordered_extracted_files_since_last_full_run(source: str) -> list[str]:
+    """Get extract files, from last full run, through all daily runs, for a source."""
+    logger.info(f"Retrieving ordered extracted files for source: '{source}'")
+    all_files = get_extracted_files_for_source(source)
+
+    # Find all full extract files and extract their dates
+    full_extracts = [f for f in all_files if _is_full_extract(f)]
+    if not full_extracts:
+        logger.warning("No full extracts found.")
+        return []
+
+    # Extract dates from full extract files and find the most recent date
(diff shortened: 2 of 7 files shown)


ehanson8 (approved):
2 optional suggestions but otherwise approved. I would caution that the `helpers` vs `utils` line may get fuzzy at times. I get not wanting to create über modules and I see the distinction you're making between essential vs. convenience but I think it forces a decision on where to put future global functions that may not always be clear. But not asking for a change, just raising the issue 🙂

abdiff/helpers/timdex_sources.py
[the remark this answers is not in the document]
jonavellecuerdo: I agree that a reordering would make the flow easier to follow. 🤔 I think we've had some conversations about the order of functions/methods (and can't recall where we landed), but I would expect: 

```
get_ordered_extracted_files_all_sources # kinda' an "entrypoint" / main function called by CLI
get_ordered_extracted_files_since_last_full_run
get_extracted_files_for_source
<everything else>
```

abdiff/helpers/timdex_sources.py
```diff
@@ -0,0 +1,93 @@
...
+            for obj in page["Contents"]:
+                if not obj["Key"].endswith("/"):  # skip folders
+                    s3_uri = f"s3://{bucket}/{obj['Key']}"
+                    files.append(s3_uri)  # noqa: PERF401
+
+    # filter where "extracted" in filename
+    return [file for file in files if "extracted" in file]
+
+
+def get_ordered_extracted_files_since_last_full_run(source: str) -> list[str]:
```
jonavellecuerdo: Hmm, what do you think about: updating `utils.parse_timdex_filename` to accept an _optional_ second argument: a list of strings denoting filename detail components (e.g. "date", "run-type"); if the second arg is provided the function returns only the specified components? You could then update the "private" methods `is_full_extract`, `is_daily_extract`, and `_extract_date` to call the util. 🤔

abdiff/helpers/timdex_sources.py
```diff
@@ -0,0 +1,93 @@
...
+        return []
+
+    # Extract dates from full extract files and find the most recent date
+    full_extract_dates = [_extract_date(f) for f in full_extracts]
+    most_recent_full_date = max(full_extract_dates)
+
+    # Collect all full extract files with the most recent date
+    most_recent_full_files = sorted(
+        f for f in full_extracts if _extract_date(f) == most_recent_full_date
+    )
```
jonavellecuerdo: Question: `most_recent_full_files` is plural because of the possibility of sequenced files (i.e., files ending with an index like "01"), right? Is there ever a scenario where we'd have: two "full" extract files for the same source on the same date (not sequenced)? 🤔

abdiff/helpers/timdex_sources.py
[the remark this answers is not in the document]
ghukill: I take your points, and agree that often the "most expensive" methods tend to drift towards the top.  However, I also think that organizing methods in the order a user or application may use them also makes sense.  Like you might see a database class with an order like:

```python
class DBClient:
  def connect():
  def close():
  def query():  
```

...where `query()` is arguably the most expensive, but conceptually also comes after managing connections.

Counter-point for this file in question: they read sequentially in order of granularity:
1. list files
2. list files, _for a single source_
3. list files, for a single source, _do so for all sources_

And where small, private functions like `_is_full_extract()` follow the first function that uses them.

I think of it almost like a file that contains functions for math, where I think a logical ordering would be:

```python
def add():
def subtract():
def multiply():
def divide():
def exponnent():
...
```

I think this file is unique, in that it likely won't change, or anything _more_ complex will build on these and thus slot to the bottom.

But, happy to reorder here, and will do so!  Just wanted to share these thoughts for future PRs, that there might be more nuance and context than just biggest and most expensive at the top.

ghukill:
> 2 optional suggestions but otherwise approved. I would caution that the `helpers` vs `utils` line may get fuzzy at times. I get not wanting to create über modules and I see the distinction you're making between essential vs. convenience but I think it forces a decision on where to put future global functions that may not always be clear. But not asking for a change, just raising the issue 🙂

Definitely take your point, and was not terribly happy with `helpers`.  I'm glad you mentioned this, as I _just_ interacted with a python library where I like a convention they use: `extras`.  It's the `psycopg2` Postgres DB driver library, and `extras` are bits and bobs that have nothing to do with core functionality, but are nice little add-ons that help you work with the library in different contexts.  Like, here, where it helps you create artifacts for running diffs... but not the diffs themselves.  

I would think then that:
- `utils` --> implies utility functions and code used by the core application
- `extras` --> functions, wrappers, data, etc. that help in specific conditions or contexts

An example of an `extra` could be something like parallel programming.  It may _wrap_ the entire application in some kind of fancy parallel programming paradigm, but absolutely not required and not part of core functionality.

Going to rename to `extras` and will re-request review, at which point please pushback if a _worse_ name!

ghukill:
@ehanson8, @jonavellecuerdo - function reordering and `helpers` to `extras` rename here: https://github.com/MITLibraries/transmogrifier-ab-diff/pull/45/commits/5701e9bd566815597042ad17f3720eb351db2f85

jonavellecuerdo (commented):
@ghukill Two small comments but looking good to me! I also ran the CLI function and it works as expected -- quick, too. :D

README.md
```diff
@@ -94,6 +94,7 @@ WEBAPP_HOST=# host for flask webapp
 WEBAPP_PORT=# port for flask webapp
 TRANSMOGRIFIER_MAX_WORKERS=# max number of Transmogrifier containers to run in parallel; default is 6
 TRANSMOGRIFIER_TIMEOUT=# timeout for a single Transmogrifier container; default is 5 hours
+TIMDEX_BUCKET=# when using CLI command 'timdex-sources-csv', this is required to know what TIMDEX bucket to use
```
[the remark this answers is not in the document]
ghukill: Not sure I totally follow... 

For running diffs we currently pass full S3 URIs, which do contain the TIMDEX bucket.  This is required now for the TIMDEX sources extra functions, which need to know what bucket to scan.

Are you suggesting that maybe in the future, we'll just pass the TIMDEX bucket, and ABDiff would just run it against _everything_?  If so -- I love the audacity -- but I'd suggest we update the docstring when that's more feasible, and more specifically when the `run-diff` argument `-i / --input-files` is no longer required.

abdiff/helpers/timdex_sources.py
ghukill: Interesting... 

While conceptually I like it, I think I might lean into this code being "extra" in the sense it shouldn't apply too much pressure for change to the "core" code.  If we had reason to extract single values from the filenames in "core" code, that'd be one thing... but I think we primarily want at least 3-4 items each time we use it.

Thoughts?

abdiff/helpers/timdex_sources.py
ghukill: Exactly.  It used to be `most_recent_full_file`!

ehanson8 (approved):
`extras` is fine by me

abdiff/helpers/timdex_sources.py
jonavellecuerdo: Ah, I think it's fine as-is! I think considering the `parse_timdex_filename` utility as part of the "core" set of functions makes sense, too.

jonavellecuerdo (approved):

PR #48, opened 2024-11-05
TIMX 383 - pipeline tweaks for large runs

ghukill:

This PR applies some updates to support large runs after some rough edges were discovered for a run ov ~1.6k input files, resulting in ~4.1m distinct records (~8.2m A/B versions).  As we get into runs this large, some of the functional-but-not-optimized code from early development needed some adjustments.

Updates include:
- [TIMX 366](https://mitlibraries.atlassian.net/browse/TIMX-366): removing transformed files after they are added to collated dataset ([commit](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/48/commits/fcd3d6e5ee3e17bc3fac1c170bdd6dfa2779423c))
  - this introduces an [optional `PRESERVE_ARTIFACTS` env var and config property](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/48/commits/fcd3d6e5ee3e17bc3fac1c170bdd6dfa2779423c#diff-a0cfcca0f834ffcd25602389082872c5dd2786e3abd9aa3f8727780338561a34R62-R66) that is used here and in later data cleanup as well; artifacts are deleted by default, but this could be helpful for debugging purposes
  - this results in a _very_ large reduction in space, where `/transformed` could be 75+ gb, but the final `/records` dataset may only be 7-8 gb
- Improve record collation ([commit](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/48/commits/6223c3ec032dfc839ed017f18b26e1338fad4cfc))
  - add logging information during intermediate work for debugging/troubleshooting
  - be more precise with columns used for intermediate CTE tables and joins
  - introduce `abdiff_record_id` as an unambiguous way to identify the A/B combination of a record throughout the run
- perform diff calculations via parallel processes ([commit](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/48/commits/e4a258b3a057c13f2a4c85920ccd703a860c8a3f))
- [TIMX 370](https://mitlibraries.atlassian.net/browse/TIMX-370): create a final "records" dataset that combines the "diffs" and "metrics" datasets ([commit](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/48/commits/1ccaa7a2e692c14c063b8d0d8f3e3f845b35bbfd))
  - webapp is updated to pull all information from "records" dataset instead of cherry picking from "difs" and "metrics" datasets

The net effect improves the creation and handling of datasets throughout the steps.  If `PRESERVE_ARTIFACTS` is not set or is `false`, then intermediate datasets will be removed once they are no longer needed.  The final, new dataset "records" contains full A/B versions for all records and the sparse matrix of modified fields, which fully supports the needs of the webapp for viewing.

The following shows an example where `PRESERVE_ARTIFACTS=true`:

```
├── 2024-11-04_19-45-24          <------------ WITH preserving artifacts
│   ├── collated
│   │   └── records-0.parquet
│   ├── diffs
│   │   └── has_diff=false
│   │       └── records-0.parquet
│   ├── metrics
│   │   └── source=libguides
│   │       └── records-0.parquet
│   ├── records
│   │   └── source=libguides
│   │       └── has_diff=false
│   │           └── records-0.parquet
│   ├── run.json
│   └── transformed
│       ├── a
│       │   ├── libguides-2024-04-03-full-transformed-records-to-index.json
│       │   └── libguides-2024-10-28-daily-transformed-records-to-index.json
│       ├── b
│       │   ├── libguides-2024-04-03-full-transformed-records-to-index.json
│       │   └── libguides-2024-10-28-daily-transformed-records-to-index.json
│       └── logs.txt
```

While the following shows if `PRESERVE_ARTIFACTS=false` or is not set (default behavior):

```
└── 2024-11-04_19-45-59           <------------ WITHOUT preserving artifacts
    ├── records
    │   └── source=libguides
    │       └── has_diff=false
    │           └── records-0.parquet
    └── run.json
```

**PR Reviewing NOTE:** There are a fair amount of files touched in this PR, but if reviewed by commit, each commit is fairly small and self-contained.  

The following will allow running a job with a fair amount of input files and records:

1- Set production AWS credentials

2- Set env var:
```shell
TIMDEX_BUCKET=timdex-extract-prod-300442551476
```

3- Init a job, produce a CSV file of input files (results in about 630), and run a diff.  **NOTE: this run will take roughly 13 minutes given the large number of input files**, but each fairly small.  This could be slightly sped up by setting `TRANSMOGRIFIER_MAX_WORKERS=9` (default is 6), but be wary this is exerting a fair amount of pressure on the system given our 10 CPU core machines.

```shell
pipenv run abdiff --verbose init-job -d output/jobs/scale -a 008e20c -b 3f77c7c

pipenv run abdiff --verbose timdex-sources-csv -o output/jobs/scale/scale.csv -s libguides,researchdatabases,gismit,gisogm,aspace

pipenv run abdiff --verbose run-diff -d output/jobs/scale -i output/jobs/scale/scale.csv
```

The final result should look similar, with only the "records" dataset persisting at the end:

```
output/jobs/scale
├── job.json
├── runs
│   └── 2024-11-05_14-50-04
│       ├── records
│       │   ├── source=aspace
│       │   │   ├── has_diff=false
│       │   │   │   └── records-0.parquet
│       │   │   └── has_diff=true                # <------ only source that had diffs for this run
│       │   │       └── records-0.parquet
│       │   ├── source=gismit
│       │   │   └── has_diff=false
│       │   │       └── records-0.parquet
│       │   ├── source=gisogm
│       │   │   └── has_diff=false
│       │   │       ├── records-0.parquet
│       │   │       └── records-1.parquet
│       │   ├── source=libguides
│       │   │   └── has_diff=false
│       │   │       └── records-0.parquet
│       │   └── source=researchdatabases
│       │       └── has_diff=false
│       │           └── records-0.parquet
│       └── run.json
```

**NOTE:** If you were to run the CLI command `view-job` at this point, when attempting to view sample records, you will run into an error along the lines of

```
duckdb.duckdb.BinderException: Binder Error: Referenced column "has_diff" not found in FROM clause!
Candidate bindings: "record_diff_matrix.related_items", "record_diff_matrix.citation"
```

This will be immediately resolved by [TIMX-385](https://mitlibraries.atlassian.net/browse/TIMX-385) (which is largely complete in local spike code) which altogether removes the idea of "samples" of records by providing a filterable table of _all_ records.  This function throwing this error is removed entirely.

#### Other notes from test run

- The joining of A and B records during `collate_ab_transforms` was formerly somewhat slow, given the number of parquest files that DuckDB would scan but discard.  This step is considerably faster.
- While not present in a dataset of this size, formerly the de-duping process could result in out of memory (OOM) errors.  This is no longer the case given an updated SQL query for de-duping in `collate_ab_transforms`.
- The calculation of A/B diffs was also a slow process, given it was performed sequentially for all records.  By running in true parallelism (multiple CPU cores), this is also considerbly quicker.

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-383
- https://mitlibraries.atlassian.net/browse/TIMX-366
- https://mitlibraries.atlassian.net/browse/TIMX-370

### Developer
- [x] All new ENV is documented in README
- [x] All new ENV has been added to staging and production environments
- [x] All related Jira tickets are linked in commit message(s)
- [x] Stakeholder approval has been confirmed (or is not needed)

### Code Reviewer(s)

[TIMX-385]: https://mitlibraries.atlassian.net/browse/TIMX-385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abdiff/core/collate_ab_transforms.py
```diff
@@ -167,89 +173,90 @@ def get_transformed_batches_iter(
...
-        ).fetch_df()["transformed_file_name"]
+        for i, transformed_file in enumerate(transformed_file_names):
+            logger.debug(
+                f"Joining records batch {i+1}/{len(transformed_file_names)}: "
+                f"{transformed_file}"
+            )
+
+            transformed_file_parquet_glob = (
+                f"{dataset_directory}/transformed_file_name={transformed_file}/*.parquet"
+            )
```
ghukill, line 250: This is one of the most key updates from this PR.

Instead of providing the full dataset of transformed files to DuckDB and then filtering on a partition of `transformed_file=XYZ`, we limit the actual parquet files DuckDB sees to this partition only.

What's key to point out: **this is NOT more memory efficient**, as DuckDB + parquet are very good at not reading data from parquet files in partitions that the WHERE clause removes, but **it is linearly faster in proportion to how many files DuckDB doesn't waste time just reading the filename**.  

This use of DuckDB + parquet is a bit unusual, as we are looping through potentially thousands of input files and performing a SQL query for each.  It's this loop + that file scan where the delay started to stack up.  If and when we use parquet in TIMDEX, we won't be performing 1-2k SQL queries in a row like this, and so even if we _do_ scan lots of files, it will be a 3-4 second pause before the query returns quickly.

abdiff/core/collate_ab_transforms.py
```diff
@@ -167,89 +173,90 @@ def get_transformed_batches_iter(
...
-                    COALESCE(a.run_type, b.run_type) run_type,
-                    COALESCE(a.action, b.action) "action",
-                    a.record as record_a,
-                    b.record as record_b
+                    uuid() as abdiff_record_id,
+                    a.timdex_record_id,
+                    a.source,
+                    a.run_date,
+                    a.run_type,
+                    COALESCE(a.action, b.action) AS action,
```
ghukill, line 270: Noting that this `COALESCE(a.action, b.action) AS action` remains non-deterministic.  @jonavellecuerdo pointed out that two different versions of Transmogrifier could _technically_ result in the same file getting interpreted as "to index" or "to delete" which would vary this `action` value.  

It is proposed that we ignore for now.  It remains a very unlikely edge case, and even more so, it actually exposes something that ABDiff is not equipped to handle yet: what if a `timdex_record_id` changes for the same intellectual record?  or in this case, the `action` changes?

This is getting into territory we should explore sometime -- what it means when Transmog varies something fundamental about the record, beyond just the metadata -- but I would propose as out of scope to ABDiff (at this time) merely showing us differences in metadata.

abdiff/core/calc_ab_diffs.py
```diff
@@ -52,29 +53,56 @@ def calc_ab_diffs(run_directory: str, collated_dataset_path: str) -> str:
...
+        for batch_count, batch in enumerate(batches_iter):
+            logger.info(f"Submitting batch {batch_count} for processing")
+            future = executor.submit(process_batch, batch)
+            pending_futures.append((batch_count, future))
+
+            if len(pending_futures) >= max_parallel_processes:
+                idx, completed_future = pending_futures.pop(0)
+                result = completed_future.result()
+                logger.info(f"Yielding diffed batch: {idx}")
+                yield result
```
ghukill, line 99: Side note: I learned something I had not considered about parallel workers during this work.

I had thought that `ProcessPoolExecutor(max_workers=X)` was sufficient to reduce CPU _and_ memory load, but that's only partially true.  While only `X` workers will actively use the CPU at once, they will be _scheduled_ all together at once, as will the arguments for each worker.  If the arguments -- data passed to the worker -- is large, this can actually put a large amount of data into memory before the worker even starts "working" on it.

As such, sometimes we need to actually throttle how quickly we even _schedule_ workers.  This code ensures that the "in flight" workers is less than a threshold, and we aren't scheduling too many at once.

Consider this example.  We have 100gb of data, but we'll only pass 1gb to each worker to perform calculations on, and we'll only have `ProcessPoolExecutor(max_workers=10)`.  I had thought this meant only 10 workers x 1gb = 10gb in memory at any given time.  But actually, we would _schedule_ 100 workers, each with 1gb of "arguments" (input data), and therefore 100gb could be stored in memory.  Sure, only 10 workers would be "working" on the 1gb they were passed each, but the waiting workers would still have their input data already in memory.  This is particularly sneaky when the input arguments are coming from an iterator; it would "drain" the iterator (the full 100gb) to schedule the workers.

What we want is actually only < 10 workers even scheduled at once, thereby slowly consuming the iterator of 100gb, and keeping memory pressure low.

This is probably poorly worded, and happy to discuss more, but worth noting.  FWIW, I have confirmed that this is memory safe as-is by following the pattern of throttling how many workers we have submitted to the `ProcessPoolExecutor()`.

ghukill: Update: I've made a personal note to revisit this.  I know for a fact that this works, and is memory safe at scale, so perhaps it's okay to proceed with it as-is at the moment.  But I'm not 100% sure what I've outlined above is accurate; I'm personally interested in digging into this a bit more, and confirming if workers can be scheduled with sizable amounts of data even before their processes are formally begun.

ehanson8 (commented):
Great updates that are clearly informed by the challenges encountered when scaling up the input! One question (that I assume has a good answer) so consider this an approval once @jonavellecuerdo gives the ✅

abdiff/core/collate_ab_transforms.py
```diff
@@ -167,89 +173,90 @@ def get_transformed_batches_iter(
...
-           the following:
-           - Create a CTE called 'transformed_file' that contains all the records
-             where the partition transformed_file_name=the name of the transformed file.
-           - Create a CTE called 'a' that contains all the records from CTE
-             'transformed_file' where the column version='a'.
-           - Create a CTE called 'b' that contains all the records from CTE
-             'transformed_file' where the column version='b'.
-           - Join the 'a' and 'b' CTEs using the column timdex_record_id.
+    As records are joined, an "abdiff_record_id" UUID is minted to unambiguously reference
+    that particular row, which is helpful later during deduping.
```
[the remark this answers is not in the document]
ghukill: Excellent question; it's not obvious at a glance, or even a medium dive.  

We use DuckDB to mint the UUID [here](https://github.com/MITLibraries/transmogrifier-ab-diff/blob/TIMX-383-pipeline-tweaks-large-runs/abdiff/core/collate_ab_transforms.py#L265) when we are joining A/B versions of records from an input file.  It's key to point out that this `timdex_record_id` (e.g. `alma:abc123`) could be in _multiple_ input files, potentially from the same day ("full" and "daily"), sometimes "to-index", sometimes "to-delete", etc.  This is why we require the de-duping stage later in the collate process.

As we move into the deduping process, we are faced with the challenge of identifying _one_ instance of an intellectual record (i.e. one A/B combination from a single input file, as records cannot duplicate within a single file.... that we know of).  We know that `timdex_record_id` is not unique across multiple files, and some edge cases have shown that even `timdex_record_id` + `run_date` + `action` may not be unique.  99% of the time that combination would be, but there are some weird edge cases.

When we dedupe, we [perform some ordering and skipping](https://github.com/MITLibraries/transmogrifier-ab-diff/blob/TIMX-383-pipeline-tweaks-large-runs/abdiff/core/collate_ab_transforms.py#L310-L335).  This addition of `abdiff_record_id` was added as part of this PR, because it was here we ran into out-of-memory (OOM) errors.  The partitioning and grouping by `timdex_record_id` required bringing _all_ rows into memory, such that we could identify the final version.  By adding the `abdiff_record_id` earlier in the process, we can [select _only_ this column during the expensive ordering operation](https://github.com/MITLibraries/transmogrifier-ab-diff/blob/TIMX-383-pipeline-tweaks-large-runs/abdiff/core/collate_ab_transforms.py#L312), dramatically reducing the memory footprint of this expensive operation.

Then, with this unambiguous identifier to that particular record version, we can [join against the full dataset](https://github.com/MITLibraries/transmogrifier-ab-diff/blob/TIMX-383-pipeline-tweaks-large-runs/abdiff/core/collate_ab_transforms.py#L346) confident we're getting only unique records.  This last join is "lazy", and is only evaluated as we consume the iterator response.

**In summary**: the addition of `abdiff_record_id` allows for less columns to select during an expensive operation, while still providing a key identifier to join later as we lazily yield those records.

This topic has come a few times over the last few months, why minting a somewhat temporary, internal identifier for ETL data can be helpful.  I think this demonstrates it pretty well.  As we chop and munge and subset data, it avoids the need to reference an intellectual "row" of data by multiple keys, that _may_ result in dupes, by providing a single, unambiguous way to join data.  In this case, it was pretty instrumental for addressing ballooning memory.

jonavellecuerdo (commented):
As discussed, here are a few questions I have for you! Thank you again for providing a lot of information on the key changes / improvements introduced in this PR. Definitely learning a lot from this project.

abdiff/core/collate_ab_transforms.py
ghukill: Previously, we would pass all parquet files to DuckDB via `.../**/*.parquet` and then rely on DuckDB to filter to the records in quesition via `WHERE transformed_file_name=XYZ`.  

DuckDB was able to keep the other files out of memory via this approach, but there was a time delay as it read the filenames + partitions from the dataset on disk to isolate the parquet files to use (and thus, the records).  So, it was memory-safe, but slower.  

If we were running 1 or 4-5 queries in a row, this wouldn't matter; the delay was small.  But when we're potentially looping through 3.2k+ (1.6k x 2, plus possible more for deletes) file names, this small delay stacks up for each query.

In the new approach, instead of DuckDB filtering the files considered for the query, we do that in advance for the glob pattern of parquet files via `transformed_file_name={transformed_file}/*.parquet`.  Now DuckDB will see like 2-3 files for it's query _and_ no need to filter them based on partition name, so much quicker.

abdiff/core/collate_ab_transforms.py
```diff
@@ -166,90 +182,99 @@ def get_transformed_batches_iter(
...
-                    COALESCE(a.run_date, b.run_date) run_date,
-                    COALESCE(a.run_type, b.run_type) run_type,
-                    COALESCE(a.action, b.action) "action",
-                    a.record as record_a,
-                    b.record as record_b
+                    uuid() as abdiff_record_id,
+                    a.timdex_record_id,
+                    a.source,
+                    a.run_date,
+                    a.run_type,
```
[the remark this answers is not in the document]
ghukill: For all these values -- `timdex_record_id`, `source`, `run_date`, and `run_type` -- we are gauranteed they will be the same value per this join.  As such, we can remove othe unecessary `COALESCE()` and just take the value from either `a` or `b`.  I missed this on our first pass on this.  

It's probably inconsequential, but at scale removing 4 `COALESCE()` calls should be a bit quicker.

abdiff/core/collate_ab_transforms.py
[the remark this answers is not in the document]
ghukill: @jonavellecuerdo - please see the comment below yours (maybe came through around the same time?), thinking perhaps this addresses your question?

abdiff/core/collate_ab_transforms.py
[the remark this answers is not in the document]
ghukill: @jonavellecuerdo - even if a record is "deleted" in version B, it would still have column values for `timdex_record_id`, `source`, `run_date`, and `run_type`.  At least it should: these can all be known from the filename of the "<source>-<date>-<run-type>-to-delete.txt" that Transmogrifier produces.  

If not, then you're right; it'd be dangerous to arbitrarily only take these columns from A or B.

And I think we have proof that these will always be set [from here](https://github.com/MITLibraries/transmogrifier-ab-diff/blob/TIMX-383-pipeline-tweaks-large-runs/abdiff/core/collate_ab_transforms.py#L138-L140) where we produce the base of the record, regardless of "to-index" or "to-delete", ensuring these values are always present.  These values come from the `filename_details`, which is produced by `parse_timdex_filename()`.

ghukill:
@jonavellecuerdo, @ehanson8 - discussions with @jonavellecuerdo revealed a couple of subtle bugs with the changes in this PR.  Going to push another commit with tweaks to address, but may push this review into next week!  Thanks for your patience and excellent feedback so far.

ghukill:
@jonavellecuerdo, @ehanson8 - the PR is ready for re-review.  @ehanson8, I know that you gave an implicit approval, but I've re-tagged you here for review given some of these recent updates.

Thanks to @jonavellecuerdo for pushing on a couple of key areas, specifically:
- the removal of `COALESCE` statements during A/B joining
- related, the correctness and reliability of `DeepDiff` usage given unusual A or B records

At this time, I still have not successfully performed a "full" run.  Given these updates I got very close over the weekend (let it run), but encountered what appears to be an intermittent Docker engine failure.  That said, I would like to continue with this PR, as I think it represents the best form so far for this application to successfully complete a "full" run (observing that a "full" run is probably not very common, but serves as a good benchmark of performance and reliability to strive for).

abdiff/core/utils.py
```diff
@@ -124,9 +129,21 @@ def write_to_dataset(
...
+
+    This function also dynamically sets option 'max_open_files' to align with current OS
+    level configurations open file limits, ensuring not to exceed, which can result in
+    "OSError: [Errno 24] Too many open files".
     """
-    written_files = []
+    # read open file limit and set max_open_files as a fraction
+    if not max_open_files:
+        soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_NOFILE)
+        max_open_files = max(int(soft_limit * 0.75), 50)
```
ghukill, line 140: This ratio of allowing max open files to be about 75% of the OS's limit is somewhat arbitrary, but based on some reading.  100% continued to throw errros, as the margins are too close.  50% worked, but is a decent performance degradation (i.e. half-speed possible of concurrent writes).  75% feels like a sweet spot for the moment of safe buffer to not exceed limit, but still pretty performant.

ehanson8 (commented):
Changes look good, one potential docstring typo. Again consider this effectively an approval once @jonavellecuerdo signs off, great work!

tests/conftest.py
```diff
@@ -561,3 +561,45 @@ def collating_intermediate_transformed_dataset(run_directory, tmp_path):
         partition_columns=["transformed_file_name"],
     )
     return transformed_dataset_filepath
+
+
+@pytest.fixture
+def mocked_transformed_files_500(run_directory):
+    """Generate 5k A and B transformed files (1k total)."""
```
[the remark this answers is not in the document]
ghukill: Yep!  Thanks, will update.

abdiff/core/utils.py
```diff
@@ -124,9 +129,21 @@ def write_to_dataset(
...
+        logger.debug(
+            f"OS open file limit of {soft_limit} detected, "
+            f"limiting max open files to {max_open_files}"
+        )
 
+    written_files = []
     ds.write_dataset(
```
jonavellecuerdo, line 147: @ghukill When is a file closed? I figured that when the max number of rows have been written to a file, then `ds.write_dataset` would close it automatically. Is that not the case? Does it wait until it hits the `max_open_files` limit before closing any files -- as this description seems to suggest:

> If greater than 0 then this will limit the maximum number of files that can be left open. If an attempt is made to open too many files then the least recently used file will be closed. If this setting is set too low you may end up fragmenting your data into many small files

jonavellecuerdo (approved):
Changes look good to me! Thank you for providing such detailed commits. ✨ Just had one question for ya'.

abdiff/core/utils.py
ghukill: The crux of the issue is partitioning.  Because we write the transformed files where each filename gets a partition, it _almost_ results in a dedicated partition (directory) for each input file, which could be thousands.

Normally, this many partitions is not ideal.  And we may even want to return to this and re-evaluate.  But it's helpful for joining, where we can limit by partition as a glob pattern.  We could also do this with individual files.

I _think_ the pyarrow dataset writer holds folder/files open for writing when it's partitioning.  If not partitioning, I think it would happily output 2k parquet files (roughly the length of the iterator for the large job).  But with partitions, I think it keeps those files open, as it may need to write to them.

I think we're close in a good architecture, but we may want to revisit this.  For example, even with the changes, it still consumes a lot of disk space.  I'm at 100+gb for transformed files only.  Granted they will get deleted as they are joined... we could theoretically performs joins right when the A and B containers are finished and avoid having those transformed files accumlate fully just to be deleted shortly thereafter.

Which is a long way of saying: **I think it's related to the partioning during writing, and I think we could/should revisit at some point.**

PR #65, opened 2024-11-13
TIMX 388, 389 - handle missing A or B records and construction of final records dataset

ghukill:

This PR addresses two outstanding tickets that prevent a successful "full" (large) run of this application.

#### [TIMX-388](https://mitlibraries.atlassian.net/browse/TIMX-388)

This was an intermittent error that was caused by incorrectly handling when either A and/or B record was missing.  As noted in various place, the A or B record can be missing for a variety of reasons, but a potentially common and valid reason is when that version of Transmogrifier cannot, or does not, produce a record.  

The bug was during calculation of metrics, when the `timdex_record_id` was identified as a "diffed" field, and therefore given a `1` or `0` (True/False) instead of the actual identifier.

The solution was:
- [more explictly handle missing A and/or B records when calculating diffs](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/65/files#diff-f6d81437fc3dd591a024246e6ea6c0b3093e76e4c16e3cb4a018736c22bf0924R137-R141)
- [add a new `a_or_b_missing` field to metrics dataset](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/65/files#diff-9080a340ec2806075e422a9e58284e7dee2a0c7ccf5f4ff2d8178c03ec5e1c97R84-R92) which can be used later if needed

#### [TIMX-389](https://mitlibraries.atlassian.net/browse/TIMX-389)

This was a pretty straight-forward out of memory (OOM) error caused by calling `to.table()` on the input "diffs" and "metrics" datasets that are joined for the final "records" dataset.  This attempts to materialize the full dataset into memory, which immediately fails.

The solution was:
- [rely on parquet globbing patterns](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/65/files#diff-8baff7c7f7c44c8342fb7278e848e1c5f2235326b2dff78f2a3be8769f2df92dR90-R108) like we do elsewhere in the application, which relies on DuckDB to cast over the parquet files and not load directly into memory

These situations are difficult to produce with small data.  I have confirmed that these updates successfully complete the steps `calc_ab_diffs` --> `calc_ab_metrics` --> `create_final_records` locally based on a "full" 91gb set of transformed files.

NO

NO

- https://mitlibraries.atlassian.net/browse/TIMX-388
- https://mitlibraries.atlassian.net/browse/TIMX-389

### Developer

### Code Reviewer(s)

[TIMX-388]: https://mitlibraries.atlassian.net/browse/TIMX-388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TIMX-389]: https://mitlibraries.atlassian.net/browse/TIMX-389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abdiff/core/create_final_records.py
```diff
@@ -38,13 +37,10 @@ def create_final_records(
...
     logger.info("Creating final records dataset from 'diffs' and 'metrics' datasets.")
     run_data = read_run_json(run_directory)
-
-    diffs_dataset = load_dataset(diffs_dataset_path)
-    metrics_dataset = load_dataset(metrics_dataset_path)
-
     metrics_timdex_field_columns = run_data["metrics"]["summary"]["fields_with_diffs"]
 
     # get list of unique columns from metrics dataset, and create final dataset schema
+    metrics_dataset = load_dataset(metrics_dataset_path)
```
jonavellecuerdo, line 43: Will this line load the full `metrics_dataset` into memory?

abdiff/core/calc_ab_metrics.py
```diff
@@ -65,22 +65,35 @@ def create_record_diff_matrix_dataset(
...
     ):
         start_time = time.time()
         batch_df = batch.to_pandas()
 
         batch_metrics = []
         for _, row in batch_df.iterrows():
+            has_diff = 1 if row["has_diff"] == "true" else 0
+            a_or_b_missing = (
+                1 if (pd.isna(row["record_a"]) or pd.isna(row["record_b"])) else 0
+            )
```
jonavellecuerdo, line 86: Hmm, do we also need to account for when `a_and_b_missing`? 🤔

abdiff/core/create_final_records.py
```diff
@@ -83,16 +80,32 @@ def create_final_records(
...
+    diffs_dataset_path: str,
+    metrics_dataset_path: str,
     metrics_timdex_field_columns: list[str],
 ) -> Generator[pa.RecordBatch, None, None]:
 
     with duckdb.connect(":memory:") as conn:
 
-        # register datasets in DuckDB for use
-        conn.register("diffs", diffs_dataset.to_table())
-        conn.register("metrics", metrics_dataset.to_table())
```
jonavellecuerdo, line 95: Or was it this line that was loading the full `diffs` and `metrics` datasets into memory? 🤔

jonavellecuerdo (commented):
Changes make sense to me but just have a few follow-up questions for you!

abdiff/core/create_final_records.py
ghukill: This is using our helper function which _lazily_ loads the dataset, so data is not pulled into memory but we can access the schema of the dataset.

abdiff/core/create_final_records.py
ghukill: Correct!   The `to.table()` was 100% to blame here.

abdiff/core/calc_ab_metrics.py
ghukill: I guess I don't totally follow?  This is where we first set that field, by setting it as true when either record a or b is missing.

jonavellecuerdo: Ah, I guess I was referring to how in `calc_ab_diffs`, we [explicitly check whether `a` **and** `b` is missing vs. `a` **or** `b` is missing -- and if we needed a column in the metrics dataset to denote that difference as well](https://github.com/MITLibraries/transmogrifier-ab-diff/pull/65/files#diff-f6d81437fc3dd591a024246e6ea6c0b3093e76e4c16e3cb4a018736c22bf0924R138-R141)?

ghukill: Yeah, I waffled a bit on the order of when we detect vs record vs handle the absence of missing record.

The approach taken here is:
- `collate_ab_transforms`
  - will happily join a record where either A and/or B is absent, doing nothing with this information
- `calc_ab_diffs`
  - the A/B records are passed as-is to sub-function `calc_record_diff`
  - this notices when A and/or B is missing, and returns the tuple `(None, set(), True|False)` depending if one or both
  - at this point, we haven't really recorded that A and/or B is missing, just that we can't really do a meaningful record-level diff
- `calc_ab_metrics`
 - here we finally set a column in a dataset, `a_or_b_missing` which indicates if one or both are missing
 - we know that this will get included in the final `records` dataset (merging of `diffs` and `metrics` datasets)

At that point, we _still_ don't do anything with this information, but the webapp could surface this in a meaningful way.  Or, we could add some more analysis [around here]() to include that in higher level summary statistics (e.g. count of records where both missing, count of records where A missing, count of records where B missing, etc.).  We could also surface on the Record page itself, but it'd be pretty obvious from looking at the A and B versions.

I had considered adding this `a_or_b_missing` at the `calc_ab_diffs` stage, or even the collating... but those would have been more invasive changes, and I wasn't seeing much benefit.  As-is, the "records-with-missing-something" kind of happily march through the steps, so they populate just as normal the datasets as they get created.  It's not until late in `calc_ab_metrics` that we finally note it, and now the decision is how do we surface that information.  But I'd propose that's work for another ticket 😎.

jonavellecuerdo: Thank you for breaking down how the flow of information between `calc_ab_diffs` and `calc_ab_metrics`! I understand it a lot better now! Your idea to include a note about missingness in "higher level summary statistics" sounds great but outside the scope of this ticket!

jonavellecuerdo (approved):

